PR #953, opened 2026-03-21
fix(development-harness): replace unsafe dict types with BacklogListItem TypedDict, fix ty diagnostics

Jamie-BitFlight:
## Summary

- **`backlog_core/operations.py`** — Extend the existing `BacklogListItem` TypedDict with `description: str` and `body: str` fields (populated by `_build_list_entry` for full-text search). Change `ListItemsResult` to `total=False` so the `**out.to_dict()` spread satisfies ty without requiring explicit `messages`/`warnings`/`errors` keys. Update `_build_list_entry` return annotation to `BacklogListItem`. Update `list_items` return annotation to `ListItemsResult`.
- **`backlog_core/server.py`** — Import `BacklogListItem` as `_BacklogListItem` under `TYPE_CHECKING` and propagate it through all item-level helper functions (`_item_field_text`, `_build_haystack`, `_item_matches_term`, all predicate `__call__` methods, `_match_body_sections`, `_collect_*`, `_enrich_with_match_context`, `_apply_search_filter`, `_dedup_by_issue_number`). Replace `dict(item)` / `dict(it)` with `{**item}` / `{**it}` for TypedDict spread compatibility. Simplify `backlog_list` items extraction to use the typed result directly. All error-handling in `_beads_lifespan`, `_assert_config`, and `_probe_backend_status` is preserved from `main`.
- **`sam_schema/cli.py`** — Validate every stdin YAML item with `_err()` (including the index of the first non-dict element) instead of silently dropping non-dict items. Validation applies to both the bare-list and `{'tasks': ...}` cases for consistency.

## Root causes

- `list_items` returned a heterogeneous `dict[str, int | list[str] | list[dict[str, str | bool]]]` with no concrete TypedDict — ty correctly rejected the unsafe narrowing. Fix: annotate with the existing TypedDicts, not cast away the error.
- All item-level helpers in `server.py` were typed as `dict[str, str | bool]`, which is incompatible with the narrower `BacklogListItem` TypedDict — fixed by propagating `_BacklogListItem` throughout.
- `ListItemsResult` declared `total=True` (default), so the `**out.to_dict()` spread could not satisfy the required `messages`/`warnings`/`errors` keys at the return site.
- Non-dict stdin YAML items were silently filtered out, potentially producing a plan with missing tasks and no diagnostic.

## Test plan

- [x] `uv run ty check plugins/development-harness/backlog_core/` — 0 errors
- [x] Pre-commit hooks passed on commit

## Follow-up

Remaining `cast()` calls in `operations.py` for GitHub GraphQL response parsing need TypedDicts — tracked separately.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -6,7 +6,7 @@
   },
   "metadata": {
     "description": "Professional development workflow extensions for Python engineers, DevOps practitioners, and AI agent developers. Covers modern Python toolchains, GitLab CI/CD automation, code quality enforcement, MCP server creation, and plugin/agent development patterns.",
-    "version": "6.3.105"
+    "version": "6.3.106"
   },
   "plugins": [
     {
--- a/plugins/development-harness/.claude-plugin/plugin.json
+++ b/plugins/development-harness/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dh",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
-  "version": "6.5.3",
+  "version": "6.5.4",
   "author": {
     "name": "Jamie Nelson",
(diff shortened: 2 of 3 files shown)


Jamie-BitFlight:
rebase this branch. @copilot and assess if the changes are still relevant

PR #2302, opened 2026-05-18
refactor(simplify): extract subprocess helper, fix silent BdInvocationError, remove dead code

Jamie-BitFlight:
Continuation of #2297 (merged). Applies cleanup to the three files introduced in that PR.

## Changes

**task_status_hook.py**
- Extract `_call_sam_fastmcp()` shared subprocess helper — `_call_sam_task_state`, `_call_sam_task_read`, `_call_sam_task_update` become thin wrappers (no API change, tests unaffected)
- Hoist `PLAN_ARG_RE` to module-level `_PLAN_ARG_RE` constant (alongside `_TASK_ID_RE`)
- Fix `_extract_plan_addr_from_path` to accept `str | Path`; remove redundant `Path()` wrapping at two call sites

**addressing.py**
- Remove redundant `ref.isdigit()` guard inside the `if ref.isdigit():` block
- Replace `int()` casts with direct string comparison — zero-padding is now significant (`P1` matches `P1-*`, not `P001-*`)

**beads.py**
- Add explicit `BdNotInstalledError` re-raise before broad `except` in `read_plan` — prevents silent empty-plan return when `bd` is absent
- Extract `_build_plan_data()` helper — eliminates duplicated 13-key `PlanData` construction
- Remove WHAT-narrating inline comments

## Test plan
- [ ] All 96 tests in `tests/test_task_status_hook.py` + `sam_schema/tests/` pass
- [ ] ty check passes
- [ ] ruff passes

### DIFF
--- a/plugins/development-harness/.claude-plugin/plugin.json
+++ b/plugins/development-harness/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dh",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
-  "version": "8.2.26",
+  "version": "8.2.27",
   "author": {
     "name": "Jamie Nelson",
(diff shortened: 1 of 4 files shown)


Jamie-BitFlight:
@codex review

PR #2625, opened 2026-06-10
fix(backlog_core): replace dict[str, object] and type:ignore with proper typed helpers

Jamie-BitFlight:
## Summary

Eliminates 3 ty warnings by fixing the underlying design rather than suppressing symptoms.

- **`test_backlog_list_divergence.py`**: `_make_list_items_result` used `dict[str, object]` as the item type, bypassing the `BacklogListItem` and `ListItemsResult` TypedDicts already defined in `operations.py`. Now imports and uses those models directly, adding the missing required fields (`plan`, `type`, `topic`) to test items and the missing `count`/`errors` fields to the returned result.

- **`test_startup_sync.py`**: Tests for non-int `GithubException.status` constructed exceptions with invalid types, requiring `# type: ignore[arg-type]`. The class comment at line 996 explicitly said "no type:ignore" but the code violated that intent. New `_make_malformed_github_exception` helper uses `GithubException.__new__` + `__dict__` injection to simulate malformed PyGitHub responses without violating the constructor's type contract.

## Design rationale

The bug in both cases is the same pattern: test code bypassed the typed domain models to use loose `dict[str, object]` / invalid constructor types. The fix strengthens types by using the models that were already there.

`_make_malformed_github_exception` uses `GithubException.__new__` + direct `__dict__` manipulation because PyGitHub stores `status` as a private name-mangled attribute and the typed constructor requires `int`. This is the only way to simulate a malformed response (where PyGitHub returns `None` or a string) without a type suppression comment.

## Test plan

- [ ] `uv run ty check plugins/` → `All checks passed!` (3 warnings → 0)
- [ ] `uv run pytest plugins/development-harness/backlog_core/tests/test_backlog_list_divergence.py plugins/development-harness/backlog_core/tests/test_startup_sync.py` → 71 passed
- [ ] All pre-commit hooks pass

https://claude.ai/code/session_01JigCULChqVEtvf1BVTL867

---
_Generated by [Claude Code](https://claude.ai/code/session_01JigCULChqVEtvf1BVTL867)_

### DIFF
--- a/plugins/development-harness/backlog_core/tests/test_startup_sync.py
+++ b/plugins/development-harness/backlog_core/tests/test_startup_sync.py
@@ -48,6 +48,7 @@ from backlog_core.sync_state import (
 if TYPE_CHECKING:
     from collections.abc import Callable
 
+    from github import GithubException
     from pytest_mock import MockerFixture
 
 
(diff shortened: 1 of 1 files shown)


Jamie-BitFlight:
@codex please review

PR #2787, opened 2026-08-05
feat: enable Codex plugin compatibility

Jamie-BitFlight:
## Summary
- add Codex plugin manifests, marketplace validation, and skill activation evidence for the Claude Skills plugin collection
- add cross-harness MCP configuration and bounded validation infrastructure
- make Development Harness project-root discovery use GitPython and remove plugin-local Python project metadata in favor of PEP 723 entrypoints
- add the project-local `what-next` skill and cross-harness research records

## Validation
- isolated Codex plugin materialization and activation validation added and exercised in the committed branch
- Development Harness standalone bundle: `uv run --script scripts/run_sam_cli.py list --plan-dir <temp> --limit 1 --format json` returned an empty JSON result without `pyproject.toml` or `uv.lock`
- Development Harness PEP 723 test runner: `46 passed` from an isolated bundle; focused repository runs: `47 passed`, `1 passed`, `1 passed`
- Ruff check and format check passed for the new PEP 723 wrappers

## Known Validation Limits
- full `prek` cannot complete on this host because root dependency resolution builds `cvxopt` and fails to locate `umfpack.h`; `skilllint` also fails because its launcher requires an unavailable `realpath` command. Targeted checks were run instead.
- worker-local coverage warnings from the root xdist configuration remain visible and were not suppressed.

Uncommitted worktree changes are not part of this PR.


Jamie-BitFlight (commented):
Automated review (code-reviewer + silent-failure-hunter sub-agents), scoped to the substantive code in this 128-file PR — the ~100 mechanical `.codex-plugin/plugin.json` manifests and `research/*.md` notes were not reviewed individually. Findings grouped below; two are critical regressions worth fixing before merge:

- **Two CI-breaking/behavior-reverting critical bugs**: a test that raises `KeyError` instead of asserting (`tests/test_frustration_analyzer_python_compatibility.py`), and `experiment-registry/server.py` reverting an already-merged fix (#2925) for resolving the wrong project root.
- **A critical import-time regression** in `dh_paths.py`: importing the module now hard-fails without a `git` binary on PATH, where previously only git-resolution call sites required it — this affects every consumer of `dh_paths`, not just Codex.
- **Evidence this PR silently reverts an already-merged commit** (#2924) via `tests/run_pytest.py` / `scripts/run_sam_cli.py`, whose branch-point predates that merge.
- A cluster of **silent-failure findings** in the new cross-harness validation/materialization scripts — swallowed stderr, uncaught JSON decode errors, batch operations with no per-item failure isolation — where a silently-passed validation would defeat the purpose of the script it's in.
- A new **required CI job for a type checker this repo has explicitly disabled** in favor of `ty`.

Everything else checked out clean: no `requests` introduced, no new `uv.lock`/workspace entries, PEP 723 headers mostly correct (one gap flagged), `ruff`/`ty` pass on touched files (one formatting gap flagged), and the core `run_bounded.py` POSIX process-group termination and `verify_cross_harness_mcp_structure.py`'s per-item error isolation are both solid reference implementations.

---

Additionally (not attachable inline — `get_task_context.py` isn't itself touched by this PR's diff, even though it's affected by the dh_paths.py import change): This script's only non-stdlib import is `dh_paths`, which as of this PR imports GitPython. This script has no `# /// script` PEP 723 header (plain `#!/usr/bin/env python3` shebang), so it now transitively requires gitpython with no declaration of that dependency — it happens to resolve under `uv run` in this checkout via the root `pyproject.toml`, but the documented direct-shebang invocation breaks on a host without gitpython installed globally. Every other DH PEP 723 script that imports `dh_paths` does declare `gitpython>=3.1.0` in its own header — this one was missed.

tests/test_frustration_analyzer_python_compatibility.py
```diff
@@ -0,0 +1,41 @@
...
+    """The server metadata includes every interpreter the plugin supports."""
+    metadata = read_pep_723_metadata(PLUGIN / "mcp" / "server.py")
+
+    assert metadata["requires-python"] == ">=3.11,<3.15"
+
+
+def test_codex_mcp_launcher_defers_interpreter_selection_to_pep_723() -> None:
+    """Codex must use the script's tested interpreter contract, not a forced minor."""
+    config = json.loads((PLUGIN / ".mcp.json").read_text(encoding="utf-8"))
+    server = config["frustration-analyzer"]
```
Jamie-BitFlight: **Critical: this will fail the `test-python` CI job as written.**

`plugins/frustration-analyzer/.mcp.json` at this PR's head nests servers under `mcpServers` (confirmed by loading the actual file: top-level keys are `['mcpServers']`), but this line does `config["frustration-analyzer"]` directly on the top-level dict — a `KeyError`, not an assertion failure. Should be `config["mcpServers"]["frustration-analyzer"]`.

plugins/scientific-method/mcp/experiment-registry/server.py
```diff
@@ -45,7 +46,7 @@ def _make_state_manager(project_root: str | None) -> StateManager:
     Returns:
         A configured ``StateManager`` instance.
     """
-    root = Path(project_root) if project_root else Path(Path.cwd())
+    root = Path(project_root) if project_root else Path(os.environ.get("PWD", Path.cwd()))
```
Jamie-BitFlight: **Critical: this reverts the fix shipped in #2925 and reintroduces the bug it fixed.**

This PR's merge-base (`f039466b`) predates `35aae026` ("fix(scientific-method): resolve experiment state to project root, not launcher cwd (#2925)"). On `main` today the root is resolved via `project_root.py::resolve_project_root()`, fed by `CLAUDE_PROJECT_DIR` (passed through in `.claude-plugin/plugin.json`'s `env` block). This branch's `plugin.json` has no `env` block and this `server.py` has no `project_root.py` — both sides bumped to overlapping versions (1.4.22/1.4.23), so merging as-is will either conflict or silently drop #2925.

`PWD` is a shell variable inherited by the launching process — stale after a `cd`, and absent/`/` when Claude is launched by an IDE or launchd — so experiment state gets written under the wrong root with no error, for every non-Codex consumer, not just the new one. This also contradicts this same PR's own `docs/codex-mcp-runtime.md`, which says the `PWD` fallback should be gated on `CODEX_THREAD_ID` and validated via GitPython — neither happens here.

Suggest rebasing onto `main` and extending `resolve_project_root()` with a Codex-gated `PWD` branch instead of replacing it.

plugins/development-harness/dh_paths.py
```diff
@@ -44,53 +45,48 @@
 from __future__ import annotations
 
 import json
+import logging
 import os
-import subprocess
 from pathlib import Path
 
+import git
```
Jamie-BitFlight, line 60: **Critical: this turns a lazy, call-site git dependency into a hard import-time failure.**

GitPython runs `refresh()` at import time and raises `ImportError: Failed to initialize: Bad git executable.` when no `git` binary is on `PATH`. Verified side by side with `PATH` stripped: on `main`, `import dh_paths` succeeds and functions that don't need git resolution (e.g. `backlog_dir(project_root=...)`) work fine; at this PR's head, the bare `import dh_paths` itself fails.

Every DH MCP server, hook, and CLI imports this module, including callers that pass an explicit `project_root` and never touch git resolution. Pre-PR, the git dependency was confined to `_git_rev_parse_root`, raising only when resolution was actually attempted. This regresses squarely against this PR's own goal — a Codex/standalone bundle without a git binary now dies at import instead of at the one function that needs git.

Fix: move `import git` inside the function that uses it (e.g. `_git_common_root`), or set `GIT_PYTHON_REFRESH=quiet` before this import.

plugins/development-harness/dh_paths.py
```diff
@@ -219,15 +233,16 @@ def infer_project_root(cwd: Path | None = None) -> Path:
...
+        return _git_common_root(start)
+    except (git.exc.InvalidGitRepositoryError, git.exc.NoSuchPathError) as exc:
         msg = (
             "Could not resolve the git project root: process cwd is not inside a repository "
             "and no project path was found. If you use Cursor, ensure the backlog MCP entry "
             "is merged from config that expands ${workspaceFolder} (see .cursor/mcp.json or "
             "plugin .mcp.json env). Otherwise set DH_PROJECT_ROOT, WORKSPACE_FOLDER_PATHS, "
             "CURSOR_PROJECT_ROOT, or CLAUDE_PROJECT_DIR, or pass --project-dir."
         )
+        _log.info("%s", msg)
```
Jamie-BitFlight: Minor: `_log.info("%s", msg)` immediately before `raise RuntimeError(msg)` emits nothing in production — this module installs no handler, so logging's lastResort floor (WARNING) swallows it. Its only observable effect is satisfying the `caplog` assertion in `test_infer_fails_with_runtime_error_when_no_project_root_is_available`, which sets the level explicitly. Also brushes against AGENTS.md's "`logging` is for debug/trace/forensic output only" — suggest deleting the call (and the caplog assertion) or demoting to `_log.debug`.

plugins/development-harness/tests/test_dh_paths.py
```diff
@@ -2,7 +2,7 @@
```
Jamie-BitFlight, line 2: `uv run ruff format --check` reports this file needs reformatting across ~22 hunks (rewrapped at ~88 columns while `pyproject.toml` sets line-length 120 — e.g. multi-line `def` signatures and parenthesized asserts that should be single lines). This is also the source of most of the unrelated churn in this file's 594-line diff, and will fail the `lint-python` CI job. Run `uv run ruff format` on it.

plugins/development-harness/tests/run_pytest.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env -S uv --quiet run --script
```
Jamie-BitFlight: This file (and the sibling `scripts/run_sam_cli.py`) already exists on `main`, added by commit `e58b9e84` (#2924) — this PR's branch point predates that merge, so `gh pr diff` shows both as new additions, but diffing this PR's version against `main`'s current version shows this branch's copy is *weaker*: the shebang drops `--active` (`uv --quiet run --script` vs. `main`'s `uv run --active --script`), and this file drops `--strict-config` entirely while downgrading `--asyncio-mode=auto` from unconditional to conditional injection. `main`'s `AGENTS.md` documents `--strict-config` as deliberate — "turns invalid or unavailable pytest configuration into a hard failure instead of a silently degraded warning." Resolving the eventual merge in favor of this branch's copy would silently revert #2924. Suggest rebasing onto `main` and dropping this file (and `run_sam_cli.py`) from this PR's diff so the already-merged version wins.

scripts/run_bounded.py
```diff
@@ -0,0 +1,92 @@
...
+        default=5.0,
+        help="Maximum command duration before process-tree termination. Default: 5.",
+    )
+    parser.add_argument("command", nargs=argparse.REMAINDER)
+    return parser
+
+
+def terminate_windows_process_tree(pid: int) -> None:
+    """Terminate a Windows process and all of its descendants."""
+    subprocess.run(["taskkill", "/PID", str(pid), "/T", "/F"], check=False)
```
Jamie-BitFlight: **Important:** this discards the result of `taskkill` entirely — no check of `returncode`, no capture/log of stderr. If `taskkill` fails (process already gone with an unexpected error, a permissions issue, or `taskkill` missing from PATH), this function returns silently as if the process tree were terminated. That directly undermines the guarantee AGENTS.md documents for this script ("terminates its full process group... including descendants") specifically on Windows — there's no verification termination actually succeeded, unlike the POSIX branch which retries with SIGKILL. Suggest checking `completed.returncode` and raising/logging (including stderr) on failure.

scripts/validate_codex_plugin_isolated.py
```diff
@@ -0,0 +1,520 @@
...
+def load_plugin_id(plugin_dir: Path) -> str:
+    """Read and validate the Codex plugin ID from a plugin manifest.
+
+    Returns:
+        The plugin identifier.
+    """
+    manifest_path = plugin_dir / ".codex-plugin" / "plugin.json"
+    if not manifest_path.is_file():
+        raise HarnessError(f"Plugin manifest is missing: {manifest_path}")
+    manifest = json.loads(manifest_path.read_text(encoding="utf-8"))
```
Jamie-BitFlight, line 178: **Important:** this `json.loads()` call (and the equivalent in `load_repo_marketplace_name()`) has no try/except, and the top-level catch at line 508 is `except (HarnessError, OSError, subprocess.SubprocessError)` — `json.JSONDecodeError` (a `ValueError` subclass) isn't in that tuple. A malformed `plugin.json` therefore escapes this script's clean `error: {err}` reporting path and crashes with a raw traceback that never names the offending file. The sibling script `validate_codex_skill_activation.py`'s top-level catch (line ~420) *does* include `json.JSONDecodeError` — worth matching that here, or wrapping each `json.loads()` to re-raise as `HarnessError(f"Malformed JSON in {manifest_path}: {error}")`.

scripts/validate_codex_skill_activation.py
```diff
@@ -0,0 +1,429 @@
...
+        except subprocess.TimeoutExpired:
+            process.kill()
+            process.wait(timeout=5)
+
+
+def run_silent(argv: list[str], *, cwd: Path, env: dict[str, str], label: str) -> None:
+    """Run a setup command without exposing its output or ambient credentials."""
+    completed = subprocess.run(argv, cwd=cwd, env=env, text=True, capture_output=True, check=False)
+    if completed.returncode != 0:
+        raise HarnessError(f"{label} failed with exit code {completed.returncode}")
```
Jamie-BitFlight: **Important:** `run_silent()` captures stdout/stderr via `capture_output=True` but on non-zero exit raises only `HarnessError(f"{label} failed with exit code {completed.returncode}")` — the actual captured stderr (the real explanation for why `codex plugin marketplace add`/`codex plugin add` failed) is read and then discarded, never surfaced. The docstring frames this as deliberate (avoiding echoing ambient credentials), which is legitimate, but the current implementation over-corrects by suppressing the entire stream rather than the credential-bearing parts, with no alternative diagnostic path. Also inconsistent with the sibling `validate_codex_plugin_isolated.py`'s `run_command()`, which prints both streams before raising. Suggest redacting known token patterns and surfacing the rest, or at minimum persisting raw stderr to the already-created evidence directory for post-mortem.

scripts/discover_cross_harness_mcp_candidates.py
```diff
@@ -0,0 +1,416 @@
...
+    """Fetch candidate metadata in GraphQL batches instead of one REST call per repo.
+
+    The existing mapping is updated in place after each completed batch, so a
+    rate-limit interruption can continue without repeating completed requests.
+    """
+    pending = [repository for repository in repositories if repository not in metadata]
+    for offset in range(0, len(pending), METADATA_BATCH_SIZE):
+        batch = pending[offset : offset + METADATA_BATCH_SIZE]
+        response = run_gh_json(["api", "graphql", "-f", f"query={build_metadata_query(batch)}"])
+        for value in response.get("data", {}).values():
```
Jamie-BitFlight, line 235: **Important:** GitHub's GraphQL API returns `null` for an aliased `repository(...)` field when a repo has been renamed, made private, or deleted since the code-search hit was recorded. That falsy `value` is silently skipped here — no counter, no warning, no recorded entry. The candidate simply vanishes from `eligible_candidates` in `rank_candidates()`, indistinguishable in the final report from a candidate correctly excluded for being below `--minimum-stars`. `collection_warnings` (built from `incomplete_queries`/`mismatched_queries`) never reflects this class of loss. Suggest tracking dropped-for-missing-metadata repos explicitly, the same way Code Search truncation is already surfaced.

scripts/sync_codex_plugin_manifests.py
```diff
@@ -0,0 +1,166 @@
...
+def load_json(path: Path) -> dict:
+    """Load a JSON object from a UTF-8 file.
+
+    Returns:
+        The decoded JSON object.
+    """
+    return json.loads(path.read_text(encoding="utf-8"))
+
+
+def sync_mcp_file(plugin_dir: Path) -> bool:
```
Jamie-BitFlight: **Important:** if `.mcp.json` exists but its `"mcpServers"` value isn't a dict (wrong shape — e.g. accidentally a list or string), both `isinstance` checks below fail and this function returns `False` — "no change needed" — with zero diagnostic that the file looks malformed; indistinguishable in output from a correctly-already-normalized file.

Separately (medium): this whole file has no `try/except` anywhere — `load_json()` (line 45) does a bare `json.loads()`, and `manifest["name"]` (line 120) is a direct index, not `.get()`. Because `main()`'s loop has no per-plugin isolation, the first plugin with malformed JSON or a missing `"name"` key aborts the *entire* repo-wide sync with an uncaught exception, and the crash message never names the offending plugin (`JSONDecodeError`'s message doesn't include the filename). For a 20+-plugin repo iterated in a `for plugin_dir in sorted(...)` loop, this means every plugin alphabetically after the bad one silently goes unprocessed. Suggest wrapping the per-plugin work in a try/except that reports the plugin directory name and either continues or aborts loudly with that context attached.

scripts/enrich_codex_plugin_manifests.js
```diff
@@ -0,0 +1,57 @@
...
+  return [...new Set(capabilities)];
+}
+
+for (const entry of fs.readdirSync(pluginsDir, { withFileTypes: true })) {
+  if (!entry.isDirectory()) continue;
+  const pluginDir = path.join(pluginsDir, entry.name);
+  const manifestPath = path.join(pluginDir, ".codex-plugin", "plugin.json");
+  if (!fs.existsSync(manifestPath)) continue;
+
+  const manifest = JSON.parse(fs.readFileSync(manifestPath, "utf8"));
```
Jamie-BitFlight: **Medium:** same pattern as the `sync_codex_plugin_manifests.py` finding on this PR — no `try/catch` anywhere in this file. A malformed `plugin.json` for any one plugin throws an uncaught `SyntaxError` mid-loop that crashes the whole batch (Node's `JSON.parse` error doesn't include the source path either), leaving every subsequent plugin directory unprocessed with no indication of what happened or how far the batch got. Suggest a `try/catch` per plugin that logs `manifestPath` and continues.

plugins/plugin-creator/scripts/auto_sync_manifests.py
```diff
@@ -738,14 +741,19 @@ def _update_from_base_ref(
     if ahead:
         return False, current_version
 
-    modified = _update_component_arrays(data, changes)
-    if modified or changes["modified"]:
+    modified = _update_component_arrays(data, changes) if sync_components else False
+    if modified or any(changes.values()):
```
Jamie-BitFlight, line 745: **Important, undocumented behavior change:** `if modified or changes["modified"]` became `if modified or any(changes.values())` here and at line 778. Because `_update_component_arrays` deliberately skips standard-path components (they're auto-discovered), an added or deleted skill/agent/command previously produced no version bump for the Claude manifest; it now does. That's arguably more correct, but it's a behavior change for every existing plugin's version-bump cadence, not just Codex manifests, and isn't mentioned in the PR body. The new test (`test_update_plugin_json_bumps_all_harness_manifests`, `tests/test_auto_sync_manifests.py:205-238`) covers the multi-manifest fan-out but only exercises the pre-existing `changes["modified"]` branch — the new `any(changes.values())` added/deleted branch this widens has no test coverage. Worth a case with `changes["added"]` populated.

.github/workflows/code-quality.yml
```diff
@@ -54,6 +54,23 @@ jobs:
       - name: Run ty
         run: uv run -q ty check .
 
+  # =============================================================================
+  # Python: basedpyright compatibility check
+  # =============================================================================
+  typecheck-basedpyright:
```
Jamie-BitFlight: **Important:** this adds `typecheck-basedpyright` as a new job that `quality-gate` requires (line 306), but `pyproject.toml` at this PR's head still carries `[tool.basedpyright] typeCheckingMode = "off"` ("Disabled: project uses ty for type checking"), unchanged by this PR. So the newly-required job is either a no-op required check or fails on unresolved imports — `uvx basedpyright` here runs in an isolated env with no project deps installed and no `uv sync` step. `AGENTS.md`'s "Type Checking" section (edited elsewhere in this PR, but not here) still states "This repository enforces **ty** (Astral) only… basedpyright are not repository quality gates." Also unpinned (`uvx basedpyright` with no version), so an upstream release could red `main` with no repo change. Per `.claude/rules/ci-workflows.md`'s Phase 5, this needs an explicit statement of what this job is actually expected to pass/fail on, and reconciling with the ty-only policy documented elsewhere.

scripts/verify_cross_harness_mcp_structure.py
```diff
@@ -0,0 +1,287 @@
...
+from operator import itemgetter
+from pathlib import Path
+from shutil import which
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+
+def run_gh_json(arguments: list[str]) -> dict[str, Any]:
```
Jamie-BitFlight: L31: `delete:` `run_gh_json` duplicated byte-for-byte (only the docstring's first line differs) from `scripts/discover_cross_harness_mcp_candidates.py:60` in this same PR. This script's own module docstring says it "consumes the ordered output of `discover_cross_harness_mcp_candidates.py`" — they're already an explicit two-stage pipeline. Replacement: `from discover_cross_harness_mcp_candidates import run_gh_json` (or extract both to a small shared `scripts/_gh_json.py`), not a second copy to keep in sync by hand.

scripts/validate_codex_plugin_isolated.py
```diff
@@ -0,0 +1,520 @@
...
+    )
+    parser.add_argument(
+        "--git-project",
+        action="store_true",
+        help="Initialize the isolated project directory as a Git repository before running Codex.",
+    )
+    return parser
+
+
+def title_case_from_kebab(name: str) -> str:
```
Jamie-BitFlight: L143: `yagni:` `title_case_from_kebab` is defined identically three times in this PR — here, `scripts/sync_codex_plugin_manifests.py:14`, and as `titleCaseFromKebab` in `scripts/enrich_codex_plugin_manifests.js:9`. Collapse the two Python copies into one shared helper; `validate_codex_plugin_isolated.py` only uses it for `--dry-run` label text, so it could even just import it.

scripts/enrich_codex_plugin_manifests.js
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env node
```
Jamie-BitFlight: `delete:` this whole file. It's never invoked from anywhere in this PR, this repo's CI, or its docs (grepped for `enrich_codex_plugin_manifests` — zero references outside this file itself), and its entire job — title-case the plugin name, detect capabilities from directory contents, write an `interface` block into the Codex manifest — is done more completely by `scripts/sync_codex_plugin_manifests.py::sync_manifest()` (line 87 in this PR), which this same PR adds as the permanent, documented, idempotent version of the same operation (also handles `skills`/`mcpServers`/`apps`/`hooks` manifest paths, which this file doesn't). The two have already drifted: this file's `detectCapabilities` checks for a `mcp` *directory* for the `Write` capability where `sync_manifest`'s Python equivalent checks for `.app.json` instead — concrete evidence of the duplication-maintenance cost, not a hypothetical one. This also reintroduces Node.js for a companion script where this repo's `.claude/rules/language-conventions.md` reserves Node.js for Claude Code hooks and routes companion scripts to Python. If this was a one-shot script used to backfill the ~20 manifests already committed in this PR, it shouldn't be committed at all (or belongs in the PR description as a note, not a permanent file); if ongoing enrichment is needed, `sync_codex_plugin_manifests.py` already owns that job. Replacement: nothing — delete the file.

Jamie-BitFlight (commented):
Second review pass — ponytail-review lens (over-engineering only, scope: unnecessary complexity, not correctness). Scoped to the substantive new code, same as the prior correctness-review pass; skipped the ~100 mechanical `.codex-plugin/plugin.json` manifests and `research/*.md` notes.

Three duplication findings, all with concrete evidence (byte-for-byte diffs / grep-confirmed zero other callers) rather than a stylistic guess — see inline comments. The three biggest new scripts (`validate_codex_plugin_isolated.py`, `validate_codex_skill_activation.py`, `discover_cross_harness_mcp_candidates.py`/`verify_cross_harness_mcp_structure.py`) are otherwise lean for what they do: the checkpoint/resume machinery, symlink-escape checks, and process-tree-safe subprocess handling are solving real problems this porting/validation effort actually has, not speculative generality — none of that is flagged.

net: -75 lines possible (delete `enrich_codex_plugin_manifests.js` entirely: -57; dedupe `run_gh_json`: ~-14; dedupe `title_case_from_kebab`: ~-6).

Issue #983, opened 2026-03-22
docs: Document skill/agent subdirectory namespace convention in plugin reference docs


Jamie-BitFlight:
## Description

Document the skill/agent subdirectory namespace convention in plugin-creator reference docs so audit agents and plugin authors know about it.

Subdirectories within skills/ or agents/ create colon-separated namespaces:
- skills/workflows/discovery/SKILL.md → invoked as /plugin:workflows:discovery
- skills/testing/foo/SKILL.md → invoked as /plugin:testing:foo

This is not documented in:
- plugin-creator:claude-plugins-reference-2026 (the primary plugin schema reference)
- plugin-creator:claude-skills-overview-2026 (the skills system reference)
- dh CLAUDE.md (lists skills with wrong invocation paths — /dh:discovery instead of /dh:workflows:discovery)

The doc-drift-auditor agent reported these skills as phantom because it checked skills/discovery/ (flat) instead of skills/workflows/discovery/ (nested). Any agent auditing plugins will make the same mistake until this convention is documented in the reference material they load.

Files to update:
1. plugins/plugin-creator/skills/claude-plugins-reference-2026/ — add subdirectory namespace section
2. plugins/plugin-creator/skills/claude-skills-overview-2026/ — add namespace convention
3. plugins/development-harness/CLAUDE.md — fix invocation paths for 10 skills (use full colon-separated paths)
4. plugins/development-harness/README.md — same fixes

### Acceptance Criteria

- [ ] `plugins/plugin-creator/skills/claude-plugins-reference-2026/` contains a "Subdirectory Warning" section stating that skill directories in subdirectories under `skills/` silently fail to register — skills must be one level deep only
- [ ] `plugins/plugin-creator/skills/claude-skills-overview-2026/` has equivalent warning with a concrete wrong/right example (e.g. `skills/testing/foo/SKILL.md` → silently dead; `skills/foo/SKILL.md` → `/plugin:foo` — correct)
- [ ] Warning text matches the canonical rule in `.claude/CLAUDE.md` lines 411-418 exactly (same constraint, same wording)
- [ ] `plugins/development-harness/CLAUDE.md` invocation paths verified flat (no `dh:group:skill` style paths) — no edits needed if already correct
- [ ] `uv run prek run --files` passes on all modified reference files with zero warnings

### RT-ICA Assessment

[...]

PR #2902, opened 2026-08-16
fix(development-harness): heal stale empty BacklogItem.reference on load

### LINKED ISSUE #2900: fix: bug(backlog-mcp): backlog groom fails with BacklogError 'Item has no backend reference' for valid open GitHub items
## Story

As a **developer relying on this plugin**, I want to **bug(backlog-mcp): backlog groom fails with backlogerror 'item has no backend reference' for valid open github items** so that **the tool works correctly and reliably**.

## Description

Observed during /dh:work-backlog-item --auto grooming of item #983 (2026-08-16). backlog view --selector "#983" and backlog update --selector "#983" --status "in-progress" both succeed and correctly resolve the item. But backlog groom --selector "#983" --section "RT-ICA" --content "..." (with and without --repo Jamie-BitFlight/claude_skills) raises:

BacklogError: Item has no backend reference
  at backlog_core/operations.py:3164 _apply_groomed_update -> "if not item.reference: raise BacklogError(...)"

Reproduced with selector as "#983", "983" (bare number), and the full title string — same failure every time, so this is not a selector-parsing issue upstream; item.reference is genuinely unset on the resolved item object at the point _apply_groomed_update runs, even though the same item resolves fine for view/update. Suggests backlog groom's item-loading path populates a different/incomplete object than backlog view/update use, or the GitHub backend's reference field (likely GraphQL node ID) isn't being carried through whatever load path groom uses.

Impact: blocks the entire grooming swarm's finalize.md write step (backlog_groom with mark_groomed=True) and any use of the RT-ICA snapshot write in groom/analyze.md — i.e. /dh:groom-backlog-item and /dh:work-backlog-item's auto-groom step (Step 3.1) cannot complete grooming on any item via the CLI in this environment. This blocked auto-mode grooming of item #983 and forced a workaround (direct implementation bypassing the SAM grooming/planning pipeline).

Repro:
uv run sam_schema/cli.py backlog groom --selector "#983" --section "RT-ICA" --content "test"
-> BacklogError: Item has no backend reference

Related: #2899 (artifact register/list write-vs-read mismatch) — filed same session, may share a root cause in how the GitHub backend resolves/persists item references after certain operations.

## Acceptance Criteria

## Context

- **Source**: Not specified
- **Priority**: P1
- **Added**: 2026-08-16

Jamie-BitFlight:
## Summary
- `groom` and `resolve` operations raised `BacklogError: Item has no backend reference` for valid, open GitHub-backed items, while `view`/`update --status` worked fine on the same items.
- Root cause: `_GitHubReconciliation.load_records()` keyed its in-memory snapshot by the raw, unhealed `BacklogItem.reference` field read straight off disk. A record persisted before `reference` was populated consistently reaches disk with `reference=""` while `metadata.issue` holds the real GitHub issue reference. `groom_item`/`_apply_groomed_update` and `resolve_item` both hard-guard on `if not item.reference: raise BacklogError(...)`. The write path (`put_work_item`) already had a fallback (`item.reference or item.issue or hash(title)`); the read path never applied the same normalization, so a stale record never self-healed and broke every subsequent groom/resolve call for that item — permanently.
- A second-order symptom of the same bug: two distinct legacy records both carrying `reference=""` would collide under the shared empty-string dict key in `load_records()`, silently dropping one from `list_work_items()`.
- Fix: factored the existing write-side fallback into a shared `_stable_reference()` helper and applied it symmetrically at the read boundary, so a stale on-disk record self-heals the moment it's loaded.

Closes #2900.

## Test plan
- [x] Added `plugins/development-harness/tests/test_github_reference_healing.py` — 7 tests covering reference healing on `list_work_items`/`get_work_item`, durable repair round-trip, collision-avoidance for two distinct stale records, `put_work_item` normalization regression guard, and end-to-end `operations.resolve_item`/`operations.groom_item` runs against a `GitHubBackend`+`FileCache` fixture reproducing the exact on-disk shape.
- [x] Falsification check: reverted the fix in a temp copy — 3 of the 5 core tests failed exactly as predicted (the two healing tests + the collision test); the 2 unrelated `put_work_item`-normalization tests still passed.
- [x] `uv run ruff check --fix`, `uv run ruff format`, `uv run ty check`, `uv run prek run --files` — all pass.
- [x] Full `plugins/development-harness/` suite: 4756 passed, 42 skipped, 5 xfailed, no failures.
- [x] Verified against this repo's real GitHub-backend item #983: `reference` on disk went from `''` to `'#983'` after one groom call.

## Notes (not addressed, out of scope)
- The offline pending-mutation queue in this repo's live cache shows `"1 failure(s)"` on every GitHub-backend reconcile — pre-existing, unrelated to the reference-healing defect. Will file separately.
- `BacklogItem.reference` is structurally disconnected from `metadata.issue` — not kept in sync by the `_sync_metadata` model validator like the other backward-compatible flat fields. This fix heals the read boundary; a deeper fix folding `reference` under the same validator-enforced invariant is a larger, more invasive model change outside this task's scope.
- `backlog update --selector ... --status "done"` silently no-ops rather than erroring (noticed during original bug investigation) — separate, unfiled concern, confirmed still present, not fixed here.

### DIFF
--- a/plugins/development-harness/backlog_core/backends/github_work_items.py
+++ b/plugins/development-harness/backlog_core/backends/github_work_items.py
@@ -422,6 +422,28 @@ class _GitHubWorkItemSync:
         return heads, comments
 
 
+def _stable_reference(item: BacklogItem) -> str:
+    """Return the opaque backend reference an item should carry.
+
+    ``BacklogItem.reference`` is not kept in sync with ``metadata.issue`` by
+    the model's ``_sync_metadata`` validator (unlike ``priority``/``issue``
+    and the other backward-compatible flat fields) — nothing guarantees the
+    two stay aligned once a record is persisted. A record written before this
+    field was populated consistently (or one round-tripped through a code
+    path that never set it) can reach disk with ``reference=""`` while
+    ``metadata.issue`` is a real GitHub issue reference. Every write already
+    self-heals through this exact fallback (see ``put_work_item``); reads
+    must apply it identically so a stale on-disk record recovers on next
+    load instead of tripping every ``reference``-gated write forever.
+
+    Returns:
+        ``item.reference`` when set, else ``item.issue``, else a stable hash
+        of the title as a last-resort opaque key for an item never linked to
+        a provider issue.
+    """
+    return item.reference or item.issue or hashlib.sha256(item.title.encode()).hexdigest()
+
+
 class _GitHubReconciliation:
     """Drive the reconciliation cycle between the private cache and the provider."""
 
@@ -459,7 +481,7 @@ class _GitHubReconciliation:
 
     def put_work_item(self, item: BacklogItem) -> None:
         """Persist a work-item intent for provider reconciliation."""
-        reference = item.reference or item.issue or hashlib.sha256(item.title.encode()).hexdigest()
+        reference = _stable_reference(item)
         self._cache._queue_work_item(reference, item.model_copy(update={"reference": reference}))
 
     def reconcile(self, request: ReconcileRequest) -> ReconcileResult:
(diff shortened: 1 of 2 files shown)


Jamie-BitFlight:
Superseded by #2916, which fixes the same root cause one layer deeper (a \`BacklogItem\` model-construction validator, per this PR's own review finding that the read-boundary patch here doesn't close the identical latent bug in \`sqlite_backend.py\`/\`memory_backend.py\`).

Verified empirically: #2916's branch heals the exact #2900 repro this PR fixes, with zero changes to \`load_records()\` — the model-level fix does it transparently at construction time for every backend. Merging both would leave two independent, redundant healing mechanisms for the same field, so closing this one rather than merging.

Thank you for the correct, well-tested fix here — it's what surfaced the deeper issue in the first place.

Issue #2963, opened 2026-08-18
fix: GitHub-backend issue-body reconcile: pending mutations accumulate and never drain across grooming sessions


Jamie-BitFlight:
## Story

As a **developer relying on this plugin**, I want to **github-backend issue-body reconcile: pending mutations accumulate and never drain across grooming sessions** so that **the tool works correctly and reliably**.

## Description

Original observation (from #2956, a different diagnosis session, 2026-08-18): across several consecutive backlog_groom writes on real GitHub-backend items (#2953, #2955), the response line "Reconciled #N: X provider patch(es), Y pending mutation(s), 0 failures" showed Y draining very slowly (29 -> 27) and never reaching 0, meaning previously-queued mutations to the linked GitHub issue's body never actually land on GitHub even though the tool reports success.

Confirmed distinct from #2956 (verified 2026-08-18 while fixing #2956): #2956's root cause is pure section-key normalization logic (operations.py/github_sync.py/rendering.py/parsing.py) with no network or sync dependency; the pending-mutation queue lives in a completely separate subsystem (backend.reconcile(), FileCache, and the GraphQL issue-update path in github_backend.py). The #2956 fix does not touch any of those files, so it cannot affect this symptom either way.

Also confirmed NOT the same mechanism as #2907 ("artifact/plan content writes never land — GitHub ruleset blocks direct Contents API commits to main"): #2907 is specifically about GitHubBackend.put_content() (artifact manifests, plans) using the Contents API. backlog_groom's issue-body updates go through a different mechanism (GraphQL updateIssue via the SyncProvider.reconcile() path in github_backend.py), so #2907's branch-protection-on-Contents-API root cause does not directly explain this symptom — needs its own trace.

**Research first**: confirm live whether GITHUB_TOKEN reaches the actual write attempt during this flow (this investigator's own fastmcp-based reproduction attempt hit an unrelated environment confound — GITHUB_TOKEN not forwarded into the fastmcp subprocess — so any repro must first rule that out, per reproduction-integrity discipline, before concluding the mutation queue itself is broken). Trace github_backend.py's reconcile()/GraphQL update path and FileCache's pending-mutation replay logic for why a queued mutation can survive a "0 failures" reconcile report without landing on GitHub. Related: #2956, #2903, #2907.

## Acceptance Criteria

- [ ] Work matches description
- [ ] Plan or implementation complete

## Context

- **Source**: Agent task - auto-derived from input, split out of #2956
[...]

Issue #2962, opened 2026-08-18
fix: view_item YAML-fallback surfaces raw section storage keys instead of display titles


Jamie-BitFlight:
## Story

As a **developer relying on this plugin**, I want to **view_item yaml-fallback surfaces raw section storage keys instead of display titles** so that **the tool works correctly and reliably**.

## Description

operations._build_sections_from_yaml_item (backlog_core/operations.py) is used by view_item's YAML-only fallback path (no rendered GitHub body available) and returns section metadata keyed by the raw internal storage key (e.g. "rt_ica", "unknown__decision") instead of the human-readable display title (e.g. "RT-ICA", "Decision") that other code paths already use via rendering.section_display_title().

Where observed: reproducible on unmodified main for canonical sections — groom an item into section "RT-ICA" with no linked GitHub issue/body, then call view_item; result.sections shows key "rt_ica", not "RT-ICA". Discovered while fixing #2956 (backlog_groom section-key duplication bug); that fix corrected _normalize_section_key to route custom section names through the same normalizer GitHub-body parsing uses, which extended this same pre-existing raw-key exposure to custom/unknown sections too (e.g. "unknown__decision" instead of "Decision").

Impact: an agent calling backlog_view for an item that has never synced to GitHub sees internal storage keys in the sections dict instead of the section names it wrote — confusing but not data-lossy (content is intact under the correct, if unreadable, key).

**Research first**: Fix likely: route _build_sections_from_yaml_item's per-key output through rendering.section_display_title(key, ...) before returning, matching the pattern _render_section_index already uses for the same data. Confirm no caller relies on the raw key as a dict key for programmatic re-lookup before changing the key shape.

## Acceptance Criteria

- [ ] Work matches description
- [ ] Plan or implementation complete

## Context

- **Source**: Agent task - auto-derived from input
- **Priority**: P2
- **Added**: 2026-08-18
- **Research questions**: None

Jamie-BitFlight:
## Description

operations._build_sections_from_yaml_item (backlog_core/operations.py) is used by view_item's YAML-only fallback path (no rendered GitHub body available) and returns section metadata keyed by the raw internal storage key (e.g. "rt_ica", "unknown__decision") instead of the human-readable display title (e.g. "RT-ICA", "Decision") that other code paths already use via rendering.section_display_title().

[...]

Issue #2922, opened 2026-08-16
fix: Pre-existing failure: test_github_content_provider_replay_preserves_concurrent_remote_revision


Jamie-BitFlight:
## Story

As a **developer relying on this plugin**, I want to **pre-existing failure: test_github_content_provider_replay_preserves_concurrent_remote_revision** so that **the tool works correctly and reliably**.

## Description

`plugins/development-harness/tests_backlog/test_github_sync_provider.py::test_github_content_provider_replay_preserves_concurrent_remote_revision` fails on current `main` (confirmed on commit `303b2319`, before and after PR #2906's `list_content()` pending-clobber fix — same assertion, same failure message: `assert 'queued' == 'revision-two'`).

Discovered by the PR #2906 fix agent while validating the `cache_content()` pending-guard fix — the full `plugins/development-harness/` suite run showed this as the only failure, and it reproduces identically on the pre-fix base commit, so it's unrelated to that PR's change.

The test's own name suggests `get_content()`'s "remote wins over a stuck pending write with a mismatched expected_revision" contract is currently broken — this is adjacent to the exact pending/CAS code in `backlog_core/backends/github_contents.py` and `file_cache.py` that PRs #2906 and #2918 both touched this session, so it could resurface in future pending-write/cache work in that area.

Source: PR #2906 fix agent's concerns block, commit cbec84fa.

## Acceptance Criteria

- [ ] Work matches description
- [ ] Plan or implementation complete

## Context

- **Source**: PR #2906 review-fix session
- **Priority**: P2
- **Added**: 2026-08-16
- **Research questions**: None

Jamie-BitFlight:
## Description

`plugins/development-harness/tests_backlog/test_github_sync_provider.py::test_github_content_provider_replay_preserves_concurrent_remote_revision` fails on current `main` (confirmed on commit `303b2319`, before and after PR #2906's `list_content()` pending-clobber fix — same assertion, same failure message: `assert 'queued' == 'revision-two'`).

Discovered by the PR #2906 fix agent while validating the `cache_content()` pending-guard fix — the full `plugins/development-harness/` suite run showed this as the only failure, and it reproduces identically on the pre-fix base commit, so it's unrelated to that PR's change.

[...]

PR #2964, opened 2026-08-18
fix(development-harness): unify section-key normalization between local writes and GitHub-parsed sections

### LINKED ISSUE #2956: fix: backlog_groom writes report success but are invisible via backlog_view — section-key duplication in _normalize_section_key plus broken GitHub sync replay
## Story

As a **developer relying on this plugin**, I want to **backlog_groom writes report success but are invisible via backlog_view — section-key duplication in _normalize_section_key plus broken github sync replay** so that **the tool works correctly and reliably**.

## Description

Every `backlog_groom` call made during a grooming pass on #2953 and #2955 (16+ calls across 6 different agents plus the orchestrator, spanning two separate GitHub-backend items) returned `"groomed_updated": true` with zero errors, but none of the written content was retrievable afterward through `backlog_view` — via CLI or MCP, `summary=true` or `summary=false`. `sections_index` continued showing only the item's original 4 sections. Cross-checked directly against the live GitHub issue (`gh api repos/.../issues/{N}`): `updated_at` predated the entire grooming session, confirming nothing reached GitHub either.

The content was not lost — it exists in the local cache file (`~/.dh/projects/{slug}/github-cache/items/issues/{N}.yaml`), but under roughly 3x the expected number of top-level section keys. Two distinct patterns observed in the cache:
1. Duplication — a section written under its correct name (e.g. `Files`, `Resources`, `Impact`, `Impact Radius`, `Dependencies`, `Priority`, `Benefits`, `Research`, `Design Intent Alignment`, `Acceptance Criteria`, `Expected Behavior`) also appears a second time under an `unknown__{slug}` key with the same content.
2. Case/format mangling with no canonical counterpart at all — `RT-ICA` was stored only as `rt_ica`, `Fact-Check` only as `fact_check`, `Issue Classification` only as `issue_classification`; none exist under their documented Title-Case/hyphenated names anywhere in the file.
3. Per-claim fact-checker writes fragmented into separate top-level sections instead of one `Fact-Check` section (`unknown__claim_1:_"..."`, `unknown__claim_3:_"..."`, etc., with the full claim text embedded in the YAML key itself).
4. Even sections nobody touched this session (`Story`, `Context`) acquired `unknown__` duplicates.

A follow-up investigation (2026-08-18) traced this to `_normalize_section_key()` in `backlog_core/operations.py` (canonical `SECTION_HEADING` map covers only 3 entries: `fact_check`, `rt_ica`, `issue_classification`; any other section name passes through unchanged on write) combined with `parse_issue_body()` in `backlog_core/github_sync.py` (any heading not in that same 3-entry map gets bucketed as `unknown__{slug}` on re-parse instead of matched back to the key it was just written under) — plus a separate, not-yet-fully-diagnosed failure in the GitHub push/replay path that means pending mutations never actually reach the live issue. Full diagnosis with file:line citations: `/private/tmp/claude-503/-Users-jamienelson-repos-claude-skills/2da56558-aee9-4992-9946-ab27b7af9cbb/scratchpad/reports/20260818-backlog-groom-visibility-bug.md`.

Impact: any item groomed through the GitHub backend via `backlog_groom` with non-canonical section names — i.e. essentially every real grooming pass — silently produces an unreadable, duplicated section structure that never reaches the actual GitHub issue, while every tool call in the chain reports success. `mark_groomed=True` would advance the item's status label while its real content stays effectively empty to any consumer. This has been running unnoticed across many prior grooming sessions.

## Acceptance Criteria

## Context

- **Source**: Session observation — discovered live while grooming #2953 via /dh:work-backlog-item, confirmed via root-cause investigation 2026-08-18
- **Priority**: P0
- **Added**: 2026-08-18

## Reopened 2026-08-19 — closure was premature, underlying defect still reproduces post-fix

Closed `completed` on 2026-08-18T08:49:47Z. Real fix commits for the cited root cause did land and are present in the currently-running plugin build (verified: installed `dh@jamie-bitflight-skills` v9.0.19, `installPath: .../dh/9.0.19`, commit `6cb65498`, byte-diffed identical against repo source; `git log 6cb65498..HEAD -- plugins/development-harness/` = 0, confirming no drift):

- `bb8ee0c0` fix(development-harness): unify section-key normalization between local writes and GitHub-parsed sections (#2964)
- `e9c1d72d` feat(development-harness): canonical section-name registry with write-boundary validation (#2987)
- `e4f4a8ba` fix(development-harness): view_item YAML-fallback shows display titles, not raw section keys (#2971)
- `c527427e` fix(development-harness): remaining PR #2987 Copilot review findings (#3015)

`backlog_core/section_registry.py` now exists with a comprehensive canonical `SectionKey` enum covering RT-ICA, Fact-Check, Issue Classification, and even Working Register (258 lines) — this is genuine, substantive progress and directly addresses the "3-entry map" gap this issue originally diagnosed in `_normalize_section_key()`.

**A fresh live reproduction (2026-08-19, same items #2953 and #2955, post-`/reload-plugins`) shows the user-visible symptom — `backlog_groom` reports success but content is unreadable via `backlog_view` and never reaches the live GitHub issue — still occurs, in a mutated form:**

- Canonical Title-Case keys DO now get created correctly for most sections (`Dependencies`, `Files`, `Impact`, `Impact Radius`, etc.) — this part is fixed.
- But a SECOND, duplicate lowercase-snake key for the *same content* also gets written alongside it (`Dependencies` + `dependencies`, `Impact Radius` + `impact_radius`, etc.) in the local cache — 29 top-level keys on #2953 where ~11 should exist.
- `RT-ICA`, `Fact-Check`, and `Issue Classification` still exist *only* as lowercase-snake (`rt_ica`, `fact_check`, `issue_classification`) with **no** canonical Title-Case/hyphenated key at all, despite `section_registry.py` defining exactly these as canonical `SectionKey` members with Title-Case/hyphenated display strings.
- Regardless of key correctness, `backlog_view(selector="#2955", section="Working Register")` — querying the exact canonical name defined in `section_registry.py` line 123 — returns `{"error":"Section not found: 'Working Register'", "valid_sections":[...only original item-creation sections...]}` immediately after a write that reported `"groomed_updated": true`. Tested fresh, isolated, with a uniquely-marked test string, specifically to rule out stale-cache artifacts.
- `gh api repos/Jamie-BitFlight/claude_skills/issues/2953` and `.../2955` — the live issues — still show zero trace of any of this session's ~20+ successful-looking writes, before this manual edit.

**Reproduction steps (exact, repeatable):**

1. Have any open GitHub-backend backlog item, e.g. `#2955`.
2. Call `mcp__plugin_dh_backlog__backlog_groom(selector="#2955", section="Working Register", append=True, content="<any unique marker string>")`. Observe response: `"groomed_updated": true`, with a `"Reconciled #2955: 1 provider patch(es), N pending mutation(s), 0 failure(s)."` message.
3. Immediately call `mcp__plugin_dh_backlog__backlog_view(selector="#2955", summary=False, section="Working Register")`.
4. Observe: `{"error":"Section not found: 'Working Register'", "valid_sections": [...only the item's original creation-time sections...]}` — despite `Working Register` being a defined canonical `SectionKey` in `section_registry.py`.
5. Cross-check ground truth: `gh api repos/Jamie-BitFlight/claude_skills/issues/2955 --jq '.body'` — the marker string from step 2 is absent.

**Conclusion**: the write-side normalization fix (#2964/#2987) is real and appears to have closed the *write* half of the original diagnosis, but the *read/render* path used by `backlog_view` (and whatever process is supposed to push local cache mutations to the live GitHub issue body) still does not surface newly-written sections at all — canonical key or not. This looks like a distinct, still-open defect downstream of the write fix, not the same bug recurring identically. `#2974`'s own reproduction (scratch item #2975, same date) found the identical "`section_filter_miss: false` but content absent" signature for the `sections=[...]` filtered-read call shape; this session confirms the same absence for the general `summary=true`/`summary=false` unfiltered read and for exact single-`section=` lookups — the read-path gap looks total, not shape-specific.

Recommend auditing the actual read/render code path `backlog_view` uses to build `sections_index`/`body` for GitHub-backend items — confirm whether it reads the local YAML cache's `sections` dict at all, or only ever renders from the last-synced GitHub issue body — and separately audit why local→GitHub push (the "pending mutations" queue, observed at 67+ and not visibly draining across many writes in this session) isn't delivering writes to the live issue.

Related: #2974 (P0 audit of silent-miss consumer impact, still open), #2970 (canonical registry — now substantially implemented via `section_registry.py`), #2573/#2504 (sync/cache architecture redesign — may be the right place to fix the push/replay half of this).

Note: this update was written directly via `gh api` PATCH to the raw issue body, bypassing `backlog_groom`/`backlog_core`'s section-write path entirely — deliberately, because that path is the one under test and a comment-only note was confirmed not to be visible to any backlog-consuming process.

## Root Cause — Diagnosed 2026-08-19 (fresh session, independent traces)

Re-reproduced the exact repro steps above against `#2955` this session (marker `REPRO-2956-TRACER-20260819-a91f7c`, since struck). Confirmed the symptom is still real, then dispatched two independent, falsification-tested traces (`dh:codebase-analyzer`, live source, live GitHub backend) — one for the write path, one for the read/push path. Full reports with confirmation/falsification tests for every claim:

- Write path: `.tmp/scratch/reports/20260819-issue-2956-write-path-trace.md`
- Read/push path: `.tmp/scratch/reports/20260819-issue-2956-read-path-trace.md`

**Headline: this issue's headline symptom has one dominant, universal root cause (A) plus one item-specific aggravating root cause (B). The write-normalization fix (#2964/#2987) is confirmed correct going forward — it is not the live defect.**

### Root cause A (dominant) — `backlog_view` reads the wrong document for every GitHub-backend item

GitHub has two body representations for a work item in this system's design: the raw human-owned `issue.body`, and the agent-owned authoritative body reachable via a `dh-content` head record → tagged audit comment (see `backlog_core/ARCHITECTURE.md` "GitHub writable records"). `backlog_groom` writes correctly land in the **second** representation — confirmed: the write posts a new issue comment via `render_work_item_comment` and CAS-writes a `WorkItemHead` record; `gh api .../issues/2955/comments` shows the comment landing at the exact write timestamp, marker included, heading `## Working Register` present.

`backlog_view`, however, unconditionally reads the **first**: `gh_client.view_enrich_from_github` does `result.body = gh_issue["body"]` — the raw GraphQL issue body — with no branch that resolves the head/comment representation `backlog_groom` actually writes to. `operations._assemble_view_content` then derives `valid_sections`/`section_filter_miss` from `## `/`### ` headings of that raw body via regex. Confirmed character-for-character: `gh api .../issues/2955 --jq '.body' | grep -E '^#{2,3} '` produces exactly the `valid_sections` array `backlog_view` returned.

Falsified independently: running the same call against live source with `GITHUB_TOKEN` unset (raw-body enrichment unavailable) correctly falls through to `item.sections` and finds `Working Register`. With the token set (normal operation), it does not. Same code, same item, same section name — only the presence of the raw-body overwrite differs.

The reconciliation path in this same codebase already resolves this correctly (`_GitHubWorkItemSync.provider_item_from_issue` calls `work_item_version()` and uses `version.body`) — only the view path's GitHub enrichment bypasses it. This affects every GitHub-backend item uniformly, independent of any section-key naming issue.

**This is why the write-normalization fix did not resolve the symptom**: the fix corrected which key a write lands under, but `backlog_view` was never reading that structure to begin with.

### Root cause B (aggravating, item-specific) — pre-registry stale keys never heal, causing real data loss on GitHub-body reparse

`#2955`'s local cache carries a stale `Working Register` (Title-Case, bare) key alongside the canonical `working_register` key. This was **not** created by the current write path — reproduced live on a clean item (`#929`, four fresh section writes including `Working Register`, `RT-ICA`, `Fact-Check`, `Issue Classification`): all four landed under exactly one canonical snake_case key each, zero duplication, zero `unknown__` keys. The stale `Working Register` key on `#2955` predates commit `bb8ee0c0` (`git log -S'"Working Register"'` confirms the name didn't exist in the registry before that commit; the stale key's entries are all timestamped before it) — it's legacy data from the pre-registry code path that returned an unresolved name verbatim.

The live defect: `rendering.normalize_unknown_sections` (the only load-time healing fold) only heals keys with an `unknown__` prefix — a bare non-canonical key like `Working Register` is treated as already-canonical and never folded into `working_register`, even though `resolve_section_name("Working Register")` resolves cleanly. `github_sync.merge_item`'s `set(local) | set(remote)` union means nothing else ever drops it either — it's permanent.

This is not cosmetic. `github_sync.render_issue_body` emits `## Working Register` twice (once for the canonical key via the entry-bearing loop, once for the bare Title-Case key via the unknown-sections loop, since `_KNOWN_SECTION_KEYS` is snake-key-only). `parsing.extract_sections` is a dict comprehension keyed on heading text — last occurrence wins. Confirmed by round-tripping `#2955`'s real cache item through the real functions: canonical section carrying 9 entries (including every recent write) renders first and is **silently discarded on reparse**, leaving only the stale 7-entry version. Same mechanism as this issue's own `#2955` Working Register entry from 2026-08-18: "`backlog_groom` writes report success but are invisible to every read path."

Blast radius (scanned all cached items): `#2953` carries 13 such recoverable bare keys (worst-affected item, matches this issue's original observation). `#2979`/`#2984`/`#2985` carry an **unrecoverable** `Description` section key — `parse_issue_body` special-cases the `## Description` heading as `item.description`, so a `Description` section key can never round-trip regardless of any registry fix; it's write-only orphan data.

The RT-ICA/Fact-Check/Issue-Classification "snake-only, no Title-Case counterpart" observation from the 2026-08-19 reproduction above is **not** a behavioral divergence — reproduced identically on `#929`: those three land snake-only today, same as `Working Register` does on a clean item. The asymmetry on `#2955`/`#2953` is registration timing (those three were the *only* three registered pre-`bb8ee0c0`, so no other name has ever had a chance to accumulate the same kind of stale sibling), not a live code-path difference.

### Root cause C (pre-existing, separate class) — pending mutations don't drain

The "pending mutation(s)" counter (66, reproduced this session; previously observed at 29 and 67+ in other sessions) sums two durable queues in `~/.dh/projects/{slug}/github-cache/cache.yaml`. For `#2955`, the queued work-item mutation is permanently stuck because `reconcile()`'s acknowledge predicate requires an exact title match between the local record and the live GitHub title, and the live title carries a `chore: ` prefix (added by `gh_client.create_issue_for_item`) that was never persisted back to the local record — this is the already-identified `#2963` class, affecting roughly 58 legacy items created before that guard existed. A separate 4-entry leak has empty `metadata.issue`. A third factor: `_reconcile_groomed_item` always uses `ReconcileScope.TARGETED` on the single item being groomed, so no groom call ever sweeps any other item's stuck mutation. **The queue count is a health-signal false alarm, not evidence that any given write failed** — every write reproduced this session landed correctly in its comment/head representation regardless of queue depth.

### Corrected understanding vs. this issue's earlier text

- "local→GitHub push... isn't delivering writes to the live issue" (line above, 2026-08-19 update) — **incorrect as stated**. Writes reach GitHub every time, in their designed representation (audit comment + head record), confirmed via `gh api .../comments`. The push/replay mechanism is not broken; `backlog_view` simply never looks at what it delivered.
- "write-normalization fix... appears to have closed the *write* half" — **confirmed true for the write path itself** (see root cause B's `#929` reproduction: zero duplication on fresh writes). The `Working Register`/`working_register` duplication visible on `#2955` today is 100% pre-existing stale data from before the fix, not new duplication — it looked like a recurrence only because nothing ever healed it.

### Fix directions (diagnosis only — nothing implemented this session)

1. **Root cause A** (highest priority — universal, blocks every consumer of `backlog_view` on every GitHub-backend item): route `gh_client.view_enrich_from_github` through the same head/comment resolution `_GitHubWorkItemSync.provider_item_from_issue` already uses, instead of the raw `gh_issue["body"]`.
2. **Root cause B**: widen `normalize_unknown_sections`'s fold gate from "key starts with `unknown__`" to "key is not a registered canonical `SectionKey` value" — this requires redefining `is_canonical_key` consistently with the existing `canonical_owns` tie-break logic per the `#3015` Copilot review note already in that function's docstring. Separately decide a policy for the unrecoverable `Description`/`Notes`/probe-key class (`#2979`/`#2984`/`#2985`/`#2900`/`#2903`) — likely a one-time migration rather than a registry entry, since `Description` can never be a section per `parse_issue_body`'s existing special-case.
3. **Root cause C**: persist the GitHub-assigned title (including any `type: ` prefix) back to the local record at issue-creation time so the title-equality guard can pass for future items; separately decide whether to write a migration/backfill for the ~58 already-stuck legacy mutations, or accept them as permanently unacknowledgeable given the queue count is not otherwise harmful to correctness.

Related: #2974 (P0 audit of silent-miss consumer impact — root cause A directly explains its `section_filter_miss: false but content absent` signature too, since that's the same raw-body vs. authoritative-body gap manifesting through a different `backlog_view` call shape), #2970 (canonical registry — root cause B confirms it's complete for the *write* path; the *read-time healing* gap is the remaining piece), #2963 (title-mismatch class, root cause C).

Test residue from this session's reproduction (marker entries on `#2955` and `#929`) has been struck via `backlog_strike_entry`.

## Fixed — 2026-08-20 (root causes A and B, greenlit by user)

Root causes A and B (from the diagnosis above) are fixed on branch `issue-2956-backlog-groom-visibility-trace`, commits `3abaf37f` and `c490ca12`. Root cause C (pending-mutation title-mismatch, ~58 legacy items) was explicitly scoped out — confirmed non-blocking for correctness, tracked as a separate follow-up rather than pulled into this fix.

### `3abaf37f` — Root cause A: `backlog_view` now reads the authoritative body

`gh_client.view_enrich_from_github` gained an optional `resolve_version` callback. `GitHubBackend.view_enrich_from_github` now wires it to `self._work_items.work_item_version()` — the same resolution `_GitHubWorkItemSync.provider_item_from_issue` already used elsewhere in this codebase — instead of the raw `gh_issue["body"]`. A resolution failure (content store unavailable, conflicting/malformed head or comment) falls back to the raw issue body rather than failing the view outright.

### `c490ca12` — Root cause B: stale non-canonical section keys now self-heal at load time

`rendering.normalize_unknown_sections` redefined "canonical" as "already a registered `SectionKey` value" (checked against `SECTION_HEADING`) instead of "does not start with `unknown__`", so a bare stale key like `Working Register` (pre-`bb8ee0c0` legacy data) now folds into its canonical `working_register` counterpart the same way an `unknown__`-prefixed key already did. Unresolvable keys (`Description`, `Notes`, the `*-PROBE` keys) are left unchanged, as before — they need a separate migration decision, not a registry fix.

Also fixed in the same commit: `_GitHubWorkItemSync.load_records()` merges cached snapshots (normalized via `yaml_io.load_item`) with queued pending mutations (loaded via plain `BacklogItem.model_validate`, which never normalized) — and a pending mutation takes precedence over its snapshot for the same reference. An item with a stuck pending mutation (the root-cause-C class) would have kept serving un-normalized sections indefinitely regardless of the registry fix above, since the healed snapshot would never be the one actually returned. Pending-mutation items are now normalized the same way before merging.

### Validation

Ran the exact reproduction steps from this issue's body against live source (direct import, not the installed plugin cache) and the live GitHub backend, on `#2955` — the item that originally exhibited both the stuck-mutation and stale-key conditions:

1. `groom_item(selector="#2955", section="Working Register", append=True, content="<marker>")` → `groomed_updated: true`, `"1 provider patch(es)"` applied.
2. `view_item(selector="#2955", section="Working Register")` immediately after → `section_filter_miss: False` (previously `True`, "Section not found"), marker present in the returned body.
3. Resolved body's `## Working Register` heading count: 1 (previously 2 — the stale bare key's duplicate render is gone).
4. Cross-checked directly against the live GitHub issue comment (`gh api .../issues/2955/comments`) — marker present.

`uv run ruff check`, `uv run ruff format --check`, and `uv run ty check` all pass on every touched file. Full `pytest` run: 1079 passed, 9 skipped, no regressions. Diagnostic test-residue entries from both the trace and validation passes were struck from `#2955` and `#929` after use.

Not in scope for this fix (tracked separately): root cause C's ~58 legacy items stuck behind the title-mismatch pending-mutation guard, and the unrecoverable `Description`/`Notes`/probe-key class on `#2979`/`#2984`/`#2985`/`#2900`/`#2903` (orphan data that needs a migration decision, not a registry entry).

## Blast-radius cleanup — 2026-08-20

Per user request, went further than the fix itself and cleaned up every item already corrupted by this bug, plus closed a third gap found during that cleanup. Commit `7922c2f4` (write-side guard), plus direct data-repair against the live GitHub backend for the affected items (not a code change — a one-time migration).

### Additional fix: `_normalize_section_key` now rejects `section="Description"`

Found live on `#2979`/`#2984`/`#2985` while scanning the local cache for remaining corruption: a `backlog_groom(section="Description", ...)` write produces a permanent, unrenderable orphan key — `github_sync.parse_issue_body` special-cases `## Description` as `BacklogItem.description`, never as a section, so no registry entry can ever make it resolvable. Same failure shape as root cause B (double-rendered heading, last-heading-wins data loss on reparse) but for a name a registry fix can't close. `_normalize_section_key` now raises a clear `BacklogError` for this name, pointing the caller at `update_item(description=...)` instead.

### Every item corrupted by this bug (found via full local-cache scan) is now clean

| Item | Corruption | Resolution |
|---|---|---|
| `#2953` | 13 bare duplicate section keys (`Priority`, `Impact Radius`, `Design Intent Alignment`, etc.) | Self-healed via `normalize` + targeted `reconcile` — fresh push, zero duplicate keys remain |
| `#2955` | `Working Register`/`working_register` duplicate | Already healed during fix validation (see above) |
| `#2975` | `Impact Radius`/`Resources` duplicates | Same normalize+reconcile pass as `#2953` |
| `#2979` | Orphan `Description` key, byte-identical to top-level `description` | Redundant copy removed, zero data loss |
| `#2984` | Orphan `Description` key, byte-identical to top-level `description` | Redundant copy removed, zero data loss |
| `#2985` | Orphan `Description` key had **diverged** from the top-level field (newer content not present elsewhere) | Unique content merged into `description` via `update_item()` before removing the orphan key — see comment on #2985 for the full account |
| `#2972` | Orphan `Notes` key (throwaway scratch-item test content) | Removed — closed scratch item, content had zero recovery value |
| `#2900` | Orphan `RT-ICA-PROBE` key (test/verification residue) | Removed — closed item, probe content |
| `#2903` | Orphan `RECONCILE-PROBE` key (debug test strings) | Removed — closed item, probe content |

Re-scanned the entire local cache (403 items) after cleanup: **zero items with a bare non-canonical section key remain.**

Every affected item was verified individually post-cleanup: `view_item()` against the live GitHub backend shows zero duplicated `##` headings in the resolved body for all nine items.

## Merged — 2026-08-19

PR #3040 merged to `main` at `6214055a`. Root causes A and B fixed, the `Description`-write guard added, full blast-radius cleanup done and verified (see above). Closing.

**Not in scope here, tracked separately:**
- Root cause C — the ~58 legacy items stuck behind the `#2963` title-mismatch pending-mutation guard. Confirmed non-blocking for correctness this session; still a real gap worth its own fix.
- A follow-up lint cleanup (pre-existing unused-argument findings in two test files, unrelated to this bug, found while addressing PR review feedback) is in progress as its own separate PR.

Jamie-BitFlight:
## Summary
- `operations._normalize_section_key` passed unrecognized section names through verbatim on write (e.g. `"Files"`), while `github_sync.parse_issue_body` normalized the same heading, once parsed back from a rendered GitHub body, to `"unknown__files"` — two different dict keys for the same section survived every merge, accumulating `unknown__` duplicates on every groom + reconcile cycle.
- Fixed by moving `heading_to_unknown_key` into `rendering.py` (the shared, backend-neutral module already hosting its inverse, `unknown_key_to_heading`) and routing `_normalize_section_key`'s fallback through it, so local writes and GitHub-parsed sections now converge on one key.
- Separately, `parsing.extract_sections` treated any `## ` line as a new section boundary even inside an entry's `<div>...</div>` content — a fact-checker verdict legitimately quoting `## Claim N: "..."` per claim shattered one `Fact-Check` entry into several spurious `unknown__claim_n...` sections on a GitHub round-trip. Fixed by tracking `<div>` nesting depth and only treating `## ` as a boundary at depth 0.
- Updated 9 existing test assertions across 5 test files that encoded the old, buggy verbatim-key expectation (e.g. `"Plan"` instead of the now-correct `"unknown__plan"`) — each individually verified via `git stash` to be a pre-existing-behavior assertion, not a new regression.
- Filed two new backlog items for issues discovered but out of scope for this PR: #2962 (a separate, pre-existing raw-key-exposure bug in `view_item`'s YAML-fallback display path, reproducible on unmodified `main`) and #2963 (the GitHub issue-body reconcile pending-mutation drain defect, confirmed architecturally distinct from this fix).

## Root cause
Two independent normalizers for "unknown" section names — one on the local-write path, one on the GitHub-parse path — disagreed on the storage key for identical section names, so a local write and its own GitHub round-trip never matched during merge.

## Test plan
- [x] Reproduced both defects directly against real corrupted cache content for issue #2953 (read-only) before fixing
- [x] Added permanent regression tests: `TestSectionKeyRoundTripRegression`, `TestExtractSectionsEntryBoundary` in `plugins/development-harness/tests/test_github_sync.py`
- [x] Re-ran the exact same reproductions after the fix — all pass
- [x] Validated against the live backlog MCP server (not just pytest) per AGENTS.md — `backlog_groom` + `backlog_view` round-trip on a throwaway scratch item shows a single `Files` section, no `unknown__files` duplicate
- [x] Full suite: `uv run pytest plugins/development-harness/tests plugins/development-harness/backlog_core/tests plugins/development-harness/tests_backlog` — 3605 passed, 1 pre-existing unrelated failure (`test_github_content_provider_replay_preserves_concurrent_remote_revision`, backlog #2922)
- [x] `uv run ruff check --fix` / `uv run ruff format` / `uv run ty check` clean on touched files

Fixes #2956

### DIFF
--- a/plugins/development-harness/backlog_core/operations.py
+++ b/plugins/development-harness/backlog_core/operations.py
@@ -73,7 +73,7 @@ from .parsing import (
     today,
     view_result_from_local_item,
 )
-from .rendering import SECTION_HEADING
+from .rendering import SECTION_HEADING, heading_to_unknown_key
 
 _SAM_SUCCESSFUL_STATUSES: frozenset[str] = _SAM_CORE_SUCCESSFUL_STATUSES | {"closed", "done"}
 _SAM_PLAN_PAGE_SIZE: Final = 100
@@ -882,23 +882,37 @@ def _apply_groomed_entries(
 
 
 def _normalize_section_key(name: str) -> str:
-    """Return the canonical snake_case storage key for a section name.
+    """Return the canonical storage key for a section name.
 
     Resolves display names (e.g. ``"RT-ICA"``) and aliases (e.g. ``"rt-ica"``) to
     the snake_case key used in ``SECTION_HEADING`` (e.g. ``"rt_ica"``).  Unknown
-    and custom section names are returned unchanged so they pass through as-is.
+    and custom section names (e.g. ``"Files"``, ``"Impact Radius"``) are routed
+    through :func:`heading_to_unknown_key` — the same normaliser
+    ``github_sync.parse_issue_body`` applies to an unrecognised ``## Heading``
+    parsed back from a GitHub issue body.
+
+    Without this, a local write of e.g. ``"Files"`` stored the key verbatim
+    while a GitHub round-trip of the same rendered heading produced
+    ``"unknown__files"``: two different dict keys for one logical section,
+    so both survived every subsequent merge and the item accumulated a
+    duplicate under ``unknown__`` on every groom + reconcile cycle (#2956).
+    Routing both call sites through the same normaliser makes local writes
+    and GitHub-parsed sections collide on one key instead of two.
 
     Lookup order:
     1. ``SECTION_HEADING_ALIAS`` keyed by ``name.lower()`` — catches hyphened aliases.
     2. Reverse scan of ``SECTION_HEADING`` for a matching display value — catches
        display names stored verbatim (e.g. ``"RT-ICA"`` → ``"rt_ica"``).
-    3. Return *name* unchanged when no match is found.
+    3. Return *name* unchanged when it is already a normalised ``unknown__`` key
+       (idempotent re-normalisation of a key read back from storage).
+    4. Otherwise, normalise via :func:`heading_to_unknown_key`.
 
     Args:
-        name: Section name as provided by the caller (e.g. ``"RT-ICA"`` or ``"rt_ica"``).
+        name: Section name as provided by the caller (e.g. ``"RT-ICA"`` or ``"Files"``).
 
     Returns:
-        Canonical snake_case key (e.g. ``"rt_ica"``), or *name* unchanged for unknown sections.
+        Canonical storage key, e.g. ``"rt_ica"`` for a canonical section or
+        ``"unknown__files"`` for a custom one.
     """
     alias_key = SECTION_HEADING_ALIAS.get(name.lower())
     if alias_key is not None:
@@ -906,7 +920,9 @@ def _normalize_section_key(name: str) -> str:
     for snake_key, display in SECTION_HEADING.items():
         if display == name:
             return snake_key
-    return name
+    if name.startswith("unknown__"):
+        return name
+    return heading_to_unknown_key(name)
 
 
 def _write_groomed_to_item(
--- a/plugins/development-harness/backlog_core/operations.py
+++ b/plugins/development-harness/backlog_core/operations.py
@@ -899,6 +899,12 @@ def _normalize_section_key(name: str) -> str:
     Routing both call sites through the same normaliser makes local writes
     and GitHub-parsed sections collide on one key instead of two.
 
+    Leading/trailing whitespace on *name* is stripped before every lookup
+    below, so e.g. ``"RT-ICA "`` (trailing space) still resolves to the
+    canonical key rather than falling through to :func:`heading_to_unknown_key`
+    and producing a key that would not match a subsequent GitHub round trip
+    of the same (already-trimmed) heading text.
+
     Lookup order:
     1. ``SECTION_HEADING_ALIAS`` keyed by ``name.lower()`` — catches hyphened aliases.
     2. Reverse scan of ``SECTION_HEADING`` for a matching display value — catches
@@ -914,6 +920,7 @@ def _normalize_section_key(name: str) -> str:
         Canonical storage key, e.g. ``"rt_ica"`` for a canonical section or
         ``"unknown__files"`` for a custom one.
     """
+    name = name.strip()
     alias_key = SECTION_HEADING_ALIAS.get(name.lower())
     if alias_key is not None:
         return alias_key
--- a/plugins/development-harness/backlog_core/operations.py
+++ b/plugins/development-harness/backlog_core/operations.py
@@ -886,7 +886,7 @@ def _normalize_section_key(name: str) -> str:
 
     Resolves display names (e.g. ``"RT-ICA"``) and aliases (e.g. ``"rt-ica"``) to
     the snake_case key used in ``SECTION_HEADING`` (e.g. ``"rt_ica"``).  Unknown
-    and custom section names (e.g. ``"Files"``, ``"Impact Radius"``) are routed
+    and custom section names (e.g. ``"Custom Analysis"``) are routed
     through :func:`heading_to_unknown_key` — the same normaliser
     ``github_sync.parse_issue_body`` applies to an unrecognised ``## Heading``
     parsed back from a GitHub issue body.
@@ -907,25 +907,29 @@ def _normalize_section_key(name: str) -> str:
 
     Lookup order:
     1. ``SECTION_HEADING_ALIAS`` keyed by ``name.lower()`` — catches hyphened aliases.
-    2. Reverse scan of ``SECTION_HEADING`` for a matching display value — catches
-       display names stored verbatim (e.g. ``"RT-ICA"`` → ``"rt_ica"``).
+    2. Case-insensitive reverse scan of ``SECTION_HEADING`` for a matching display
+       value — catches display names stored verbatim (e.g. ``"RT-ICA"`` → ``"rt_ica"``),
+       matching case-insensitively for parity with the parse-side lookup
+       (``github_sync.py``'s ``_HEADING_TO_KEY.get(heading.lower())``) so a caller-supplied
+       ``"story"`` normalises the same as GitHub-parsed ``"Story"`` instead of falling
+       through to :func:`heading_to_unknown_key`.
     3. Return *name* unchanged when it is already a normalised ``unknown__`` key
        (idempotent re-normalisation of a key read back from storage).
     4. Otherwise, normalise via :func:`heading_to_unknown_key`.
 
     Args:
-        name: Section name as provided by the caller (e.g. ``"RT-ICA"`` or ``"Files"``).
+        name: Section name as provided by the caller (e.g. ``"RT-ICA"`` or ``"Custom Analysis"``).
 
     Returns:
         Canonical storage key, e.g. ``"rt_ica"`` for a canonical section or
-        ``"unknown__files"`` for a custom one.
+        ``"unknown__custom_analysis"`` for a custom one.
     """
     name = name.strip()
     alias_key = SECTION_HEADING_ALIAS.get(name.lower())
     if alias_key is not None:
         return alias_key
     for snake_key, display in SECTION_HEADING.items():
-        if display == name:
+        if display.lower() == name.lower():
             return snake_key
     if name.startswith("unknown__"):
         return name
--- a/plugins/development-harness/backlog_core/parsing.py
+++ b/plugins/development-harness/backlog_core/parsing.py
@@ -741,39 +741,24 @@ def extract_description_from_issue_body(body: str) -> str:
 def extract_sections(text: str) -> dict[str, str]:
     """Extract '## Section' content blocks from markdown text.
 
-    A ``## `` line is only treated as a section boundary while scanning is
-    outside every ``<div>...</div>`` entry block. Entry content (see
-    ``entry_blocks.py``) is free-form and may legitimately contain its own
-    ``## `` headings (e.g. a fact-checker verdict quoting one claim per
-    heading) — without this guard those embedded headings are indistinguishable
-    from real section boundaries and shatter one section into many spurious
-    ones on the next parse.
+    Delegates the actual boundary detection to :func:`_split_body_h2`, the
+    marko-AST-based splitter already used by the legacy ``.md`` body-parsing
+    path (see the "Legacy .md body section parsing" section of this module).
+    Both callers now share one structural section boundary decision instead
+    of independently maintaining their own hand-rolled line scanners that can
+    silently disagree — see :class:`_EntryDivBlock` for why a naive line-based
+    ``## `` scan (this function's previous implementation) misidentifies a
+    heading-shaped line inside entry-block content as a real section boundary.
 
     Args:
         text: Markdown body text.
 
     Returns:
         Dict mapping section heading (e.g. '## Story') to its content (not including heading).
+        When the same heading text appears more than once, the last occurrence wins —
+        matching this function's pre-existing (dict-based) contract.
     """
-    sections: dict[str, str] = {}
-    current_heading: str | None = None
-    current_lines: list[str] = []
-    div_depth = 0
-
-    for line in text.splitlines():
-        if div_depth == 0 and line.startswith("## "):
-            if current_heading is not None:
-                sections[current_heading] = "\n".join(current_lines).strip()
-            current_heading = line.strip()
-            current_lines = []
-        elif current_heading is not None:
-            current_lines.append(line)
-        div_depth = max(0, div_depth + line.count("<div>") - line.count("</div>"))
-
-    if current_heading is not None:
-        sections[current_heading] = "\n".join(current_lines).strip()
-
-    return sections
+    return {f"## {name}": content for name, content in _split_body_h2(text)}
 
 
 def merge_sections(local_body: str, github_body: str) -> tuple[str, bool]:
@@ -829,7 +814,11 @@ def merge_sections(local_body: str, github_body: str) -> tuple[str, bool]:
 # ---------------------------------------------------------------------------
 
 import marko
-from marko.block import Heading as _MarkoHeading
+from marko.block import BlockElement as _MarkoBlockElement, Heading as _MarkoHeading
+from marko.helpers import MarkoExtension as _MarkoExtension
+
+if TYPE_CHECKING:
+    from marko.source import Source as _MarkoSource
 
 # Heading levels used for body section splitting.
 _H2_LEVEL = 2
@@ -840,6 +829,84 @@ _GROOMED_DATE_RE = re.compile(r"^Groomed(?:\s*\((\d{4}-\d{2}-\d{2})\))?$", re.IG
 
 # ATX heading detector used when mapping AST positions back to source lines.
 _ATX_HEADING_RE = re.compile(r"^(#{1,6})\s+(.*)")
+
+# Entry-block wrapper markers. Must match entry_blocks.wrap_entry()'s
+# "<div><sub>{ts}</sub>\n\n{content}\n</div>" format (see ENTRY_RE in
+# entry_blocks.py for the canonical parse-side pattern this mirrors).
+_ENTRY_DIV_OPEN = "<div><sub>"
+_ENTRY_DIV_CLOSE = "</div>"
+_ENTRY_DIV_OPEN_RE = re.compile(r" {0,3}" + re.escape(_ENTRY_DIV_OPEN))
+
+
+class _EntryDivBlock(_MarkoBlockElement):
+    """Marko block element matching this codebase's ``<div><sub>...</sub>...</div>`` entry wrapper.
+
+    CommonMark's HTML-block rules end a generic ``<div>`` block at the first
+    blank line. ``entry_blocks.wrap_entry()`` always emits a blank line right
+    after the opening ``<div><sub>{ts}</sub>`` tag (to separate the timestamp
+    from multi-paragraph content), so marko's built-in ``HTMLBlock`` element
+    would otherwise stop treating the entry as opaque at that first blank line
+    and reparse everything after it as ordinary markdown — including any
+    ``## ``/``### ``-looking line inside the entry's own content (e.g. a
+    fact-checker verdict quoting one claim per heading) as a real section
+    boundary (#2956).
+
+    This element is registered with higher priority than ``Heading``,
+    ``HTMLBlock``, and ``Paragraph`` (see :data:`_ENTRY_AWARE_MARKDOWN`), so
+    once a line opens an entry block the parser hands the *entire* block —
+    every line up to the point ``<div>``/``</div>`` nesting returns to zero,
+    regardless of blank lines, code fences, or heading-shaped text inside — to
+    this element instead of descending into it looking for block-level
+    structure. Depth is tracked (not "stop at the first ``</div>``") because
+    entry content may itself contain further, unrelated ``<div>``/``</div>``
+    text; a first-match stop would end the opaque region early and let a
+    heading-lookalike line *after* that inner close fragment the section
+    again. That makes "a heading-lookalike line inside entry content is
+    mistaken for a section boundary" structurally impossible rather than
+    special-cased: no line inside an entry block is ever offered to the
+    block-level parser as a heading candidate in the first place, at any
+    nesting depth.
+    """
+
+    priority = 9  # Above ThematicBreak(8)/FencedCode(7)/Heading(6)/HTMLBlock(5)/Paragraph(1).
+
+    def __init__(self, lines: str) -> None:
+        self.body = lines
+
+    @classmethod
+    def match(cls, source: _MarkoSource) -> bool:
+        """Return whether the current line opens an entry block.
+
+        Returns:
+            ``True`` when the line matches the entry-block open marker.
+        """
+        return bool(source.expect_re(_ENTRY_DIV_OPEN_RE))
+
+    @classmethod
+    def parse(cls, source: _MarkoSource) -> str:
+        """Consume every line until ``<div>``/``</div`` nesting returns to zero.
+
+        Returns:
+            The raw, verbatim source text of the entry block.
+        """
+        lines: list[str] = []
+        depth = 0
+        while not source.exhausted:
+            line = source.next_line()
+            if line is None:
+                break
+            lines.append(line)
+            source.consume()
+            depth += line.count("<div>") - line.count("</div>")
+            if depth <= 0:
+                break
+        return "".join(lines)
+
+
+# Single shared marko instance used by every AST-based split in this module so
+# entry-block opacity is enforced identically everywhere, rather than each
+# caller independently re-deciding what counts as a section boundary.
+_ENTRY_AWARE_MARKDOWN = marko.Markdown(extensions=[_MarkoExtension(elements=[_EntryDivBlock])])
 
 
 def _extract_heading_text(node: _MarkoHeading) -> str:
@@ -861,7 +928,8 @@ def _extract_heading_text(node: _MarkoHeading) -> str:
 def _ast_h2_headings(text: str) -> list[str]:
     """Return level-2 heading texts from the marko AST in document order.
 
-    marko correctly ignores ``##`` lines inside fenced code blocks.
+    Uses :data:`_ENTRY_AWARE_MARKDOWN` so ``##`` lines inside fenced code
+    blocks or entry-block content are never misidentified as headings.
 
     Args:
         text: Raw markdown body text.
@@ -869,7 +937,7 @@ def _ast_h2_headings(text: str) -> list[str]:
     Returns:
         Ordered list of heading text strings (after ``## ``).
     """
-    doc = marko.parse(text)
+    doc = _ENTRY_AWARE_MARKDOWN.parse(text)
     return [
         _extract_heading_text(child)
         for child in doc.children
@@ -886,7 +954,7 @@ def _ast_h3_headings(text: str) -> list[str]:
     Returns:
         Ordered list of heading text strings (after ``### ``).
     """
-    doc = marko.parse(text)
+    doc = _ENTRY_AWARE_MARKDOWN.parse(text)
     return [
         _extract_heading_text(child)
         for child in doc.children
@@ -897,9 +965,13 @@ def _ast_h3_headings(text: str) -> list[str]:
 def _map_headings_to_lines(ast_headings: list[str], raw_lines: list[str], target_level: int) -> list[tuple[int, str]]:
     """Map AST heading texts to their 0-indexed source line numbers.
 
-    Uses a code-fence state tracker to skip ``#`` lines inside fenced code
-    blocks, then matches AST heading entries in document order by level.  Line
-    numbers are 0-indexed (offsets into ``raw_lines``).
+    The AST decides *which* lines are real headings; this function only
+    locates *where* each one lives in the original source so its exact
+    verbatim content (including entry-block HTML) can be sliced out. Its
+    state tracking must therefore recognise the same block boundaries the
+    AST does — code fences and entry blocks — or it can misalign a heading
+    text with the wrong source line once any ``#``-prefixed text appears
+    inside one of those opaque regions.
 
     Args:
         ast_headings: Heading texts extracted from the marko AST, in order.
@@ -909,15 +981,21 @@ def _map_headings_to_lines(ast_headings: list[str], raw_lines: list[str], target
     Returns:
         List of ``(line_index, heading_text_from_source)`` tuples.
     """
-    "#" * target_level + " "
     result: list[tuple[int, str]] = []
     ast_idx = 0
     in_fence = False
+    entry_depth = 0
 
     for idx, line in enumerate(raw_lines):
         if ast_idx >= len(ast_headings):
             break
         stripped = line.lstrip()
+        if entry_depth > 0:
+            entry_depth += line.count("<div>") - line.count("</div>")
+            continue
+        if _ENTRY_DIV_OPEN_RE.match(line):
+            entry_depth = line.count("<div>") - line.count("</div>")
+            continue
         if stripped.startswith(("```", "~~~")):
             in_fence = not in_fence
             continue
--- a/plugins/development-harness/backlog_core/parsing.py
+++ b/plugins/development-harness/backlog_core/parsing.py
@@ -834,7 +834,6 @@ _ATX_HEADING_RE = re.compile(r"^(#{1,6})\s+(.*)")
 # "<div><sub>{ts}</sub>\n\n{content}\n</div>" format (see ENTRY_RE in
 # entry_blocks.py for the canonical parse-side pattern this mirrors).
 _ENTRY_DIV_OPEN = "<div><sub>"
-_ENTRY_DIV_CLOSE = "</div>"
 _ENTRY_DIV_OPEN_RE = re.compile(r" {0,3}" + re.escape(_ENTRY_DIV_OPEN))
 
 # Tag-boundary matchers for div-nesting depth tracking. A literal-substring count
--- a/plugins/development-harness/backlog_core/parsing.py
+++ b/plugins/development-harness/backlog_core/parsing.py
@@ -741,6 +741,14 @@ def extract_description_from_issue_body(body: str) -> str:
 def extract_sections(text: str) -> dict[str, str]:
     """Extract '## Section' content blocks from markdown text.
 
+    A ``## `` line is only treated as a section boundary while scanning is
+    outside every ``<div>...</div>`` entry block. Entry content (see
+    ``entry_blocks.py``) is free-form and may legitimately contain its own
+    ``## `` headings (e.g. a fact-checker verdict quoting one claim per
+    heading) — without this guard those embedded headings are indistinguishable
+    from real section boundaries and shatter one section into many spurious
+    ones on the next parse.
+
     Args:
         text: Markdown body text.
 
@@ -750,15 +758,17 @@ def extract_sections(text: str) -> dict[str, str]:
     sections: dict[str, str] = {}
     current_heading: str | None = None
     current_lines: list[str] = []
+    div_depth = 0
 
     for line in text.splitlines():
-        if line.startswith("## "):
+        if div_depth == 0 and line.startswith("## "):
             if current_heading is not None:
                 sections[current_heading] = "\n".join(current_lines).strip()
             current_heading = line.strip()
             current_lines = []
         elif current_heading is not None:
             current_lines.append(line)
+        div_depth = max(0, div_depth + line.count("<div>") - line.count("</div>"))
 
     if current_heading is not None:
         sections[current_heading] = "\n".join(current_lines).strip()
--- a/plugins/development-harness/backlog_core/parsing.py
+++ b/plugins/development-harness/backlog_core/parsing.py
@@ -837,6 +837,28 @@ _ENTRY_DIV_OPEN = "<div><sub>"
 _ENTRY_DIV_CLOSE = "</div>"
 _ENTRY_DIV_OPEN_RE = re.compile(r" {0,3}" + re.escape(_ENTRY_DIV_OPEN))
 
+# Tag-boundary matchers for div-nesting depth tracking. A literal-substring count
+# (``line.count("<div>")``) misses an attributed opening tag like
+# ``<div class="note">`` while ``line.count("</div>")`` still matches its close
+# unconditionally — that asymmetry drives depth negative and ends entry-block
+# opacity early, letting a heading-lookalike line further down escape as a
+# spurious section (#2964 follow-up). Matching tag boundaries via regex keeps
+# opens and closes counted symmetrically regardless of attributes.
+_DIV_OPEN_TAG_RE = re.compile(r"<div\b")
+_DIV_CLOSE_TAG_RE = re.compile(r"</div\s*>")
+
+
+def _div_depth_delta(line: str) -> int:
+    """Return the net change in ``<div>``/``</div>`` nesting depth for one line.
+
+    Args:
+        line: Source line to scan.
+
+    Returns:
+        Count of opening ``<div`` tags minus closing ``</div>`` tags on this line.
+    """
+    return len(_DIV_OPEN_TAG_RE.findall(line)) - len(_DIV_CLOSE_TAG_RE.findall(line))
+
 
 class _EntryDivBlock(_MarkoBlockElement):
     """Marko block element matching this codebase's ``<div><sub>...</sub>...</div>`` entry wrapper.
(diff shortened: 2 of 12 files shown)


Jamie-BitFlight:
## Rework: replaced heuristic scanner with marko AST parsing

Per review feedback, the `extract_sections` fix has been reworked. Summary:

**What changed:**
- The `<div>`-depth counter added to `extract_sections` in the first commit was a hand-rolled heuristic — a second one bolted onto `parsing.py`'s existing line-scanning approach, which this repo's `.claude/CLAUDE.md` prohibits ("Do NOT write regex parsers for markdown"). Reverted.
- `parsing.py` already had a working marko-AST section splitter (`_split_body_h2`/`_ast_h2_headings`/`_map_headings_to_lines`) for the legacy `.md` body-parsing path. `extract_sections` now **delegates to it** instead of maintaining a second, independently-written scanner — eliminating the duplicate implementation entirely, not adding a third one.
- That existing marko infrastructure turned out to share the *exact same* undiscovered bug: verified empirically that marko's built-in `HTMLBlock` stops treating a `<div>` as opaque at the first blank line (CommonMark rule), and this codebase's own `entry_blocks.wrap_entry()` always emits a blank line right after the opening tag — so even a correct AST-based approach still misidentified heading-shaped lines inside entry content as real headings, and `_map_headings_to_lines`'s raw-line remapper had the same blind spot independently. Fixed both, once, at the shared root: a custom marko `BlockElement` (`_EntryDivBlock`) teaches marko itself that this codebase's entry wrapper is one opaque block (balanced `<div>`/`</div>` depth tracking, not "stop at first `</div>`" — verified a naïve first-match stop lets a nested unrelated div re-expose a heading-lookalike line after it). This makes the bug class structurally impossible rather than special-cased.
- Added `TestSectionRoundTripProperty`, a hypothesis property test asserting write→render→parse→merge is lossless for arbitrary section names and adversarial content (embedded headings, nested divs, code fences).

**What the property test found (and how it was handled):**
- One genuinely in-scope defect, fixed directly: `_normalize_section_key`/`heading_to_unknown_key` didn't strip whitespace before deriving a key, causing the same write-path/parse-path mismatch class #2956 targets, just triggered by trailing whitespace instead of an unrecognized name.
- Three real but narrow, out-of-scope defects — filed, not fixed here: #2966 (Unicode title-casing round-trip), #2967 (`entry_blocks.ENTRY_RE` misparsing nested entry-lookalike content), #2968 (underscore/space key-encoding ambiguity).

**Audit requested in review (content_normalizer.py, disclosure_handler.py, ordinal_mapper.py):** found the same duplicate-independent-implementation pattern — filed as #2969 (3 independent generators of the `## Sections [N] Title (M entries)` format feed 1 independent parser; ordinal-path generation and validation are also independently authored). Not confirmed as currently live, not a cheap fix, so filed rather than expanding this PR.

**Validation:** full suite (3607 passed, 1 pre-existing unrelated failure), ruff, ty all clean on touched files. Re-verified against the live backlog MCP server (not just pytest) with the original fact-checker multi-claim reproduction — single `Fact-Check` section, no fragmentation.

Jamie-BitFlight:
## Registry extension + generality test (per follow-up request)

Extended `rendering.SECTION_HEADING` from 3 to 24 entries, sourced from a grep of the real corrupted cache for #2953/#2955 (ground truth) plus every literal `section=` value in `plugins/development-harness/agents/*.md` and skill references — not a partial/guessed list. Added: Files, Resources, Impact, Impact Radius, Dependencies, Priority, Benefits, Research, Design Intent Alignment, Acceptance Criteria, Expected Behavior, Effort, Reproducibility, Story, Context, Working Register, Suggested Location, Concerns, Divergence Notes, Execution Results, Grooming Notes, Root-Cause Analysis.

**This is display-quality only** — the round-trip fix in the earlier commits already makes ANY section name (registered or not) converge on one key. Added `test_unregistered_section_write_then_github_round_trip_preserves_content`, using a heading deliberately excluded from the registry, to prove this generality explicitly (not just implied by the hypothesis property test). Updated 8 existing assertions across 5 files that had used a now-registered name as their "unregistered" example.

**New finding, filed not fixed (#2973, P1):** while sourcing the observed-headings list, found that `BacklogItem.description` is rendered as bare unwrapped text — unlike every section, which is entry-div-wrapped and protected by #2956's fix. #2955's actual real description contains its own `## ` prose headings (documenting its own workflow), and reproduced directly: this still gets silently truncated/fragmented on a GitHub round trip today. Not fixed here — the only clean fix (wrapping description in the same entry-div convention) has real backward-compatibility blast radius across ~2900+ existing GitHub issues in this repo, a design decision out of scope for this PR.

Full suite (3608 passed, 1 pre-existing unrelated failure), ruff, ty clean. Re-verified rendering against the real #2953 item via the live MCP server (read-only `backlog_view`, #2953/#2955 left untouched).

Jamie-BitFlight (commented):
Automated review (code-reviewer + silent-failure-hunter sub-agents), verified against the live `.dh` backlog cache where noted. One critical migration gap, two important correctness bugs in the new entry-block depth tracking / key normalization, plus two minor suggestions. The core marko-AST entry-boundary fix and the `heading_to_unknown_key` unification themselves are solid and well covered by the new property-based tests.

---

Additionally (not attachable inline — the line isn't part of this PR's diff): **Suggestion: a third, still-un-unified key derivation.** `parse_md_body_sections` derives `raw_key = heading_name.lower()` with no space→underscore step: `## Impact Radius` → `'impact radius'` vs. the write path's `impact_radius`; `## Custom Analysis` → `'custom analysis'` vs. `unknown__custom_analysis`. Pre-existing, not a regression from this PR, but it's a live path (`.claude/backlog/*.md` currently has 174 `## Story`, 3 `## Impact Radius`, 3 `## Files` headings) and sits squarely inside the write-path/parse-path invariant this PR establishes elsewhere. Worth routing through the shared normalizer, or filing alongside #2966-#2968 as explicitly out of scope.

plugins/development-harness/backlog_core/rendering.py
```diff
@@ -30,11 +31,48 @@
...
+# through the generic unknown__ fallback (#2956's write-path/parse-path fix
+# makes that true for ANY section name, registered or not) — registering
+# them here is a display-quality improvement only, not a correctness fix:
+# it gives grooming's most commonly observed sections a clean canonical
+# storage key (e.g. "files") instead of the "unknown__" prefix. Sourced from
+# a full grep of the real corrupted local cache for #2953/#2955 (the
+# ground-truth evidence for #2956) plus every literal `section=` value found
+# across plugins/development-harness/agents/*.md and skill references
+# (2026-08-18) — not a guessed or partial list.
 SECTION_HEADING: dict[str, str] = {
```
Jamie-BitFlight, line 53: **Critical: this expansion silently re-keys ~350 existing cached items and reproduces the exact #2956 duplication this PR fixes.**

Section keys are persisted (`FileCache._serialize_item` dumps `item.model_dump()` to `~/.dh/projects/{slug}/github-cache/items/issues/*.yaml`), not re-derived. Checked the live cache: it holds `unknown__story` ×352, `unknown__context` ×352, `unknown__acceptance_criteria` ×348, `unknown__impact_radius` ×25, `unknown__files` ×22 — all newly-canonical names added here. Post-merge, `parse_issue_body` maps `## Story` → `story` (now canonical), so the cached `unknown__story` and a freshly-parsed `story` become two keys for one logical section. Verified against the actual reconcile path (`reconciliation.py:150-158`):

```
MERGED KEYS: ['unknown__story', 'story']
=> rendered body contains "## Story" twice, with the identical entry duplicated
```

It stabilizes at 2× rather than growing unbounded, but every affected issue permanently doubles that section's content on its next reconcile.

`backlog_normalize` does not help here — `operations.normalize_items` is `list_work_items()` → `put_work_item(item)` verbatim, no re-keying (grepped for migrate/legacy/rekey around `unknown__`: zero hits).

Suggest either (a) dropping these ~20 new registry entries from this PR — per your own comment two lines up, they're "a display-quality improvement only, not a correctness fix," so they carry all of the migration risk here and none of the actual #2956 fix — or (b) shipping a one-off re-key pass (`unknown__{k}` → `{k}` when `k in SECTION_HEADING`) in `normalize_items` before this registry lands.

plugins/development-harness/backlog_core/parsing.py
```diff
@@ -831,6 +830,84 @@ def merge_sections(local_body: str, github_body: str) -> tuple[str, bool]:
...
+        """
+        lines: list[str] = []
+        depth = 0
+        while not source.exhausted:
+            line = source.next_line()
+            if line is None:
+                break
+            lines.append(line)
+            source.consume()
+            depth += line.count("<div>") - line.count("</div>")
```
Jamie-BitFlight: **Important: depth counting breaks on any `<div>` carrying attributes.**

`line.count("<div>")` won't match `<div class="note">`, but `line.count("</div>")` matches its close regardless — so an attributed div inside entry content drives `depth` negative and ends the opaque region early. Verified: entry content containing `<div class="note">inner</div>` causes `extract_sections` to return `['## Fact-Check', '## Claim 1: fake heading', '## Resources']` — the `## Claim 1` line escapes as a spurious section, exactly the failure mode this class exists to prevent (and which the docstring below claims is "structurally impossible … at any nesting depth").

`wrap_entry` itself only ever emits bare `<div><sub>`/`</div>`, so the generator side is safe — the risk is entirely from agent-authored entry *content* quoting HTML with attributes, which is routine.

Fix is one line: match tag boundaries instead of literal substrings, e.g. `len(re.findall(r"<div\b", line))` / `len(re.findall(r"</div\s*>", line))`. Same fix needed at the mirrored counter in `_map_headings_to_lines` (line 994).

plugins/development-harness/backlog_core/operations.py
```diff
@@ -882,31 +882,54 @@ def _apply_groomed_entries(
...
-        Canonical snake_case key (e.g. ``"rt_ica"``), or *name* unchanged for unknown sections.
+        Canonical storage key, e.g. ``"rt_ica"`` for a canonical section or
+        ``"unknown__files"`` for a custom one.
     """
+    name = name.strip()
     alias_key = SECTION_HEADING_ALIAS.get(name.lower())
     if alias_key is not None:
         return alias_key
     for snake_key, display in SECTION_HEADING.items():
         if display == name:
```
Jamie-BitFlight: **Important: this reverse scan is case-sensitive while every other normalizer in the pair is case-insensitive.**

`_normalize_section_key("Story")` → `story`, but `_normalize_section_key("story")` → `unknown__story` (falls through to the fallback since `"story" != "Story"` here). The parse side is case-insensitive both ways (`_HEADING_TO_KEY.get(heading.lower())`, `github_sync.py:47`) — so this is the same write-path/parse-path divergence this PR exists to close, still live for any case-variant caller. Currently latent (every literal `section=` value in `plugins/development-harness/{agents,skills}` is exact Title Case), but cheap to close now: `if display.lower() == name.lower():`.

plugins/development-harness/backlog_core/operations.py
```diff
@@ -882,31 +882,54 @@ def _apply_groomed_entries(
 
 
 def _normalize_section_key(name: str) -> str:
-    """Return the canonical snake_case storage key for a section name.
+    """Return the canonical storage key for a section name.
 
     Resolves display names (e.g. ``"RT-ICA"``) and aliases (e.g. ``"rt-ica"``) to
     the snake_case key used in ``SECTION_HEADING`` (e.g. ``"rt_ica"``).  Unknown
-    and custom section names are returned unchanged so they pass through as-is.
+    and custom section names (e.g. ``"Files"``, ``"Impact Radius"``) are routed
```
Jamie-BitFlight: Minor: this docstring's examples (`"Files"`, `"Impact Radius"`) are now stale — both got registered in `SECTION_HEADING` by the later commit in this PR, so they resolve at step 2 (clean canonical key) and never reach `heading_to_unknown_key` any more. Swap in a genuinely unregistered example, e.g. `"Custom Analysis"`, as the new tests already do.

plugins/development-harness/backlog_core/parsing.py
```diff
@@ -831,6 +830,84 @@ def merge_sections(local_body: str, github_body: str) -> tuple[str, bool]:
 # ATX heading detector used when mapping AST positions back to source lines.
 _ATX_HEADING_RE = re.compile(r"^(#{1,6})\s+(.*)")
 
+# Entry-block wrapper markers. Must match entry_blocks.wrap_entry()'s
+# "<div><sub>{ts}</sub>\n\n{content}\n</div>" format (see ENTRY_RE in
+# entry_blocks.py for the canonical parse-side pattern this mirrors).
+_ENTRY_DIV_OPEN = "<div><sub>"
+_ENTRY_DIV_CLOSE = "</div>"
```
Jamie-BitFlight: L837: delete: `_ENTRY_DIV_CLOSE` constant defined but never referenced anywhere else in this file — the depth counters at L900 and L994 both hardcode the literal `"</div>"` directly instead of using it. Delete the constant, or actually route both counters through it.

Jamie-BitFlight (commented):
ponytail-review pass (over-engineering only — correctness bugs already covered in a separate review round, not re-flagged here). Three findings, all small.

Not flagged, judged proportionate: `_EntryDivBlock`'s ~20-line docstring is explaining a genuinely non-obvious design choice (balanced-depth tracking vs. first-`</div>`-stop) that the PR's own tests prove matters — not padding. `TestSectionRoundTripProperty`'s Hypothesis machinery is targeted at the exact two bug classes this PR fixes (200 examples, adversarial fragments drawn from the real failure modes) — proportionate test infrastructure for a silent-data-loss bug class, not bloat.

`net: -1 to -4 lines possible.`

plugins/development-harness/backlog_core/parsing.py
Jamie-BitFlight: L900 & L994: shrink: identical `<div>`/`</div>` depth-counting logic (`line.count("<div>") - line.count("</div>")`) duplicated verbatim in `_EntryDivBlock.parse` and `_map_headings_to_lines`. One shared one-liner — `def _div_depth_delta(line: str) -> int: return line.count("<div>") - line.count("</div>")` — called from both, removes the copy-paste-drift risk. (This is the same duplication shape that let a bug slip into one copy while the other stayed unaffected, per the separate correctness-review pass.)

plugins/development-harness/backlog_core/rendering.py
Jamie-BitFlight, line 53: L50-76: yagni: ~20 new `SECTION_HEADING` entries bundled into a bug-fix PR. The comment two lines above this dict already concedes these are "a display-quality improvement only, not a correctness fix" for #2956 — nothing about closing that issue requires them. Split into a separate PR: it's also the sole source of the critical cache-migration/re-keying risk flagged in the correctness-review pass, so keeping it out of the actual fix commit makes both easier to review, land, and revert independently.

Jamie-BitFlight: **Retracting this suggestion — the premise is wrong.** Checked whether these registrations are actually cosmetic by grepping consumers instead of trusting the PR's own "display-quality improvement only" framing:

- `skills/work-backlog-item/references/workflows/groom/groom-drift.md:81-83` extracts file paths for drift detection via `sections["Impact Radius"]` and `sections["Files"]` — both are among the ~20 entries added here.
- `skills/work-backlog-item/references/workflows/work/feasibility-gate.md:23` gates a **BLOCKED** decision (effort/priority mismatch) on `sections['Priority']` — also newly registered here.

Both are exact-key lookups against the display title this PR's `SECTION_HEADING` registry produces (via `_section_display_title`, made display-keyed by the companion PR #2971). Before this PR, those consumers were silently reading from a key that didn't exist for locally-groomed items (verbatim `"Priority"`/`"Impact Radius"`/`"Files"` never round-tripped to a display title that matched). So these entries aren't optional polish — they're what makes two real, already-written workflow gates find their data at all. Splitting them into a separate PR, as I suggested, would leave `feasibility-gate.md`'s BLOCKED check and `groom-drift.md`'s drift classification broken in the interim.

The critical migration-risk finding in the sibling comment on this same line still stands — if anything this raises its stakes, since the re-keying now affects gates that make blocking decisions, not just display. The actionable fix is the re-key/migration path (or dropping just the ~350-item risk via a one-off `normalize_items` pass), not removing the registry entries.

Jamie-BitFlight: Confirmed — this is already handled, no bulk migration/rekey pass needed.

`normalize_unknown_sections()` (commit 3ec8f365, wired into `yaml_io.py`'s `load_item`/`load_item_text`) folds any `unknown__{key}` entry into `{key}` at load time whenever `{key}` is registered in `SECTION_HEADING` — generic against whatever the registry currently holds, not hardcoded to specific names. That's exactly the "normalize on read" mechanism the design decision on #2970 calls for.

Traced and verified the two consumers you named directly. Both `groom-drift.md`'s Mode B and `feasibility-gate.md` line ~23 call `backlog view --selector "{title}"` — a title selector, which means `parse_issue_selector` returns `None` and GitHub-body enrichment never runs. For that code path, `ViewItemResult.sections` is built by `operations._build_sections_from_yaml_item`, which echoes the raw in-memory `BacklogItem.sections` key verbatim (snake_case, e.g. `impact_radius`) — not a Title-Case display string. Added two regression tests chaining the real functions those consumers exercise (`load_item` → `_build_sections_from_yaml_item`) confirming a stale `unknown__impact_radius`/`unknown__files`/`unknown__priority` cache entry now resolves under the clean `impact_radius`/`files`/`priority` key, same as a freshly-groomed item.

One nuance worth flagging: neither before nor after this fix does the *literal* Title-Case key `sections["Impact Radius"]` (as written in `groom-drift.md`'s prose) appear in that response — it's always the snake_case internal key. That mismatch predates this PR and is independent of `SECTION_HEADING` registration; nothing here changes it either way. If `groom-drift.md`/`feasibility-gate.md` are meant to do a literal Python-style dict lookup on that exact string, they'd need a separate fix (snake_case key, or a display-name lookup helper) — tracking that as a follow-up, out of scope for #2964 since it's pre-existing and touches workflow prose, not backend key derivation.

`SECTION_HEADING` keeps all ~20 entries in this PR (reverted the earlier split). Also folded the draft candidate-list rationale into #2970's issue body (comments aren't read by the grooming workflow's field extraction) rather than leaving it as a comment.

Jamie-BitFlight: Follow-up after verifying the composition directly (scratch-merged #2971's tip onto this branch, ran the regression tests, then discarded — nothing pushed here).

Your bottom-line ask (keep the registry, normalize-on-read handles storage-key consistency) is correct — confirmed again. One correction to my last reply: I said `ViewItemResult.sections` for a title-selector `backlog view` call is keyed by the raw snake_case internal key, never the Title-Case display string. That's accurate for *this* branch alone, but PR #2971 (issue #2962) changes `_build_sections_from_yaml_item` to key the response by `_section_display_title(sec_name)` instead of the raw key — so once #2971 lands, the response *is* keyed by Title Case.

Verified the full chain empirically:
1. Stale cache holds `unknown__impact_radius`.
2. This branch's `normalize_unknown_sections()` (3ec8f365) folds it to `impact_radius` at YAML-load time.
3. #2971's `_build_sections_from_yaml_item` change then keys the response by `_section_display_title("impact_radius")` → `"Impact Radius"`.
4. `groom-drift.md`'s `sections["Impact Radius"]` lookup succeeds.

Confirmed with the scratch merge: without #2971, `_build_sections_from_yaml_item` returns `{'impact_radius': ...}` (my regression tests pass on that key). With #2971 merged in, the same function returns `{'Impact Radius': ...}` (my regression tests then fail on the snake_case assertion but the Title-Case key resolves correctly instead — checked directly). Neither branch is sufficient alone: this one without #2971 would still be reading correctly-folded `impact_radius` but `groom-drift.md`'s literal lookup wouldn't match it; #2971 without this one would key the response by `_section_display_title("unknown__impact_radius")` (whatever `heading_to_unknown_key`'s reverse produces for that), not `"Impact Radius"`.

So: the two PRs are complementary, not redundant — this one fixes what key gets stored, #2971 fixes what key gets displayed in the response. `groom-drift.md`/`feasibility-gate.md` work correctly only once both merge, in either order.

Jamie-BitFlight: One more follow-up — verified the scope question you raised (whether `_build_sections_metadata`'s body-based path, which already Title-Cases via `## Heading` parsing, means real GitHub-synced items mostly avoid the `_build_sections_from_yaml_item` snake_case/raw-key issue).

Traced it precisely: `BacklogItem` — the Pydantic model persisted to the local `.yaml` cache — has no `body` field at all (checked its full field list: `title`, `reference`, `description`, `metadata`, `sections`, plus a few excluded runtime-only fields). The raw GitHub markdown (`ProviderItem.body`, a *different* model used only during a live fetch) is parsed into structured `sections` and then discarded — never retained locally. `result.body` in `operations.view_item` can only become non-empty via `view_enrich_from_github` (gated on an issue-number selector, `parse_issue_selector(selector)` truthy) or a legacy `.md` file read in `view_result_from_local_item`. Neither applies to a title selector against a `.yaml`-cached item.

So the routing in `_assemble_view_content` — `if body: ... elif item and item.sections: _populate_yaml_item_content(...)` — takes the YAML-fallback branch (`_build_sections_from_yaml_item`) for **every** title-selector `backlog view` call against any item that's ever synced from GitHub, not a narrow "created purely as local YAML" edge case. This is the common case, not the exception.

Also confirmed #2980 (which I filed) is a straight duplicate of #2962 — same function, same root cause, same fix in #2971. Updated #2980's body to reflect that and recommend closing it as a duplicate once #2971 merges.

Net: your original framing holds without qualification — #2964 + #2971 together are both required, and together they cover the broad case (any GitHub-synced item viewed by title), not a narrow slice of it.

PR #2971, opened 2026-08-18
fix(development-harness): view_item YAML-fallback shows display titles, not raw section keys

### LINKED ISSUE #2962: fix: view_item YAML-fallback surfaces raw section storage keys instead of display titles

Jamie-BitFlight:
## Summary
- `_build_sections_from_yaml_item` (used by `view_item`'s YAML-only fallback path, when an item has no rendered GitHub body) returned section metadata keyed by the raw internal storage key (e.g. `"rt_ica"`, `"unknown__decision"`) instead of the human-readable display title (`"RT-ICA"`, `"Decision"`) that every other section-display path (`## Sections` index, `_render_section_index`) already produces via `rendering.section_display_title()`.
- Root cause: `_build_sections_from_yaml_item` used `item.sections`' storage keys directly as its own output keys instead of routing them through the display-title resolver already used elsewhere in this module.
- Fix: key the returned dict by `_section_display_title(sec_name, groomed_date)` instead of the raw `sec_name`, for both `Section` and `GroomedData` branches.

## Root cause
Reproducible on unmodified `main`: groom an item into section `"RT-ICA"` with no linked GitHub issue/body, then call `view_item` — `result.sections` shows key `"rt_ica"`, not `"RT-ICA"`. Discovered while fixing #2956 (`backlog_groom` section-key duplication bug); that fix corrected `_normalize_section_key` to route custom section names through the same normalizer GitHub-body parsing uses, which surfaced this same pre-existing raw-key exposure for custom/unknown sections too (e.g. `"unknown__decision"` instead of `"Decision"`).

## Test plan
- [x] Reproduced directly against real function calls (`add_item` + `groom_item` + `view_item`, memory backend, no GitHub) on unmodified `main` before fixing
- [x] Added a permanent regression test: `test_view_item_yaml_fallback_shows_display_title_for_canonical_section` in `tests/test_backlog_core_operations.py`
- [x] Updated the existing `TestViewItemUnknownSections` class (6 tests) — these tests previously asserted the *old, buggy* contract (raw `unknown__`/canonical keys surfacing verbatim); individually inspected and confirmed each was pre-existing-behavior encoding, not a masked regression, before updating to expect display titles
- [x] Re-ran the exact reproduction after the fix — confirmed passing
- [x] Validated against the live backlog MCP server (not just pytest) per AGENTS.md — `backlog_groom` "RT-ICA" then `backlog_view` shows `"RT-ICA"`, not `"rt_ica"`
- [x] Full suite: `uv run pytest plugins/development-harness/tests plugins/development-harness/backlog_core/tests plugins/development-harness/tests_backlog` — 3600 passed, 1 pre-existing unrelated failure (`test_github_content_provider_replay_preserves_concurrent_remote_revision`, backlog #2922)
- [x] `uv run ruff check --fix` / `uv run ruff format` / `uv run ty check` clean on touched files

Fixes #2962

### DIFF
--- a/plugins/development-harness/backlog_core/operations.py
+++ b/plugins/development-harness/backlog_core/operations.py
@@ -1996,6 +1996,12 @@ def _build_sections_from_yaml_item(item: BacklogItem) -> dict[str, SectionEntryM
     for sec_name, sec_data in item.sections.items():
         if isinstance(sec_data, GroomedData):
             title = _section_display_title(sec_name, sec_data.date)
+            if title in result:
+                msg = (
+                    f"_build_sections_from_yaml_item collision: display title {title!r} already "
+                    f"holds a non-groomed section; cannot merge GroomedData into it"
+                )
+                raise TypeError(msg)
             result[title] = GroomedSectionMetadata(type="groomed", date=sec_data.date, subsections=sec_data.subsections)
         elif isinstance(sec_data, Section):
             entries = sec_data.entries
@@ -2005,7 +2011,17 @@ def _build_sections_from_yaml_item(item: BacklogItem) -> dict[str, SectionEntryM
             active_count = sum(1 for e in entries if not e.struck)
             struck_count = sum(1 for e in entries if e.struck)
             title = _section_display_title(sec_name)
-            result[title] = _SectionMetadata(num_entries=active_count, num_struck=struck_count, entries=entry_dicts)
+            if title in result:
+                existing = result[title]
+                if not _is_section_entry_metadata(existing):
+                    msg = (
+                        f"_build_sections_from_yaml_item collision: display title {title!r} already "
+                        f"holds a GroomedData section; cannot merge a Section into it"
+                    )
+                    raise TypeError(msg)
+                result[title] = _merge_section_entries(existing, entry_dicts)
+            else:
+                result[title] = _SectionMetadata(num_entries=active_count, num_struck=struck_count, entries=entry_dicts)
     return result
 
 
--- a/plugins/development-harness/backlog_core/operations.py
+++ b/plugins/development-harness/backlog_core/operations.py
@@ -1977,20 +1977,26 @@ def _build_sections_from_yaml_item(item: BacklogItem) -> dict[str, SectionEntryM
     Used when the item has no raw markdown body (i.e. it was created as a ``.yaml``
     file) but its ``sections`` field carries structured ``Section`` objects.
 
+    Keys are display titles (via :func:`_section_display_title`), not the raw
+    storage keys in ``item.sections`` — a canonical section is stored under a
+    snake_case key (e.g. ``"rt_ica"``) and a custom one under an ``unknown__``
+    key (e.g. ``"unknown__decision"``); callers of this function (``view_item``'s
+    YAML-only fallback path) expect the same human-readable names that the
+    GitHub-body and ``## Sections`` index paths already show (#2962).
+
     Args:
         item: The BacklogItem whose sections to convert.
 
     Returns:
-        Mapping of section name to entry metadata.  ``Section`` entries follow the
+        Mapping of display title to entry metadata.  ``Section`` entries follow the
         :class:`SectionEntryMetadata` shape.  ``GroomedData`` entries follow the
         :class:`GroomedSectionMetadata` shape (``"type": "groomed"`` discriminator).
     """
     result: dict[str, SectionEntryMetadata | GroomedSectionMetadata] = {}
     for sec_name, sec_data in item.sections.items():
         if isinstance(sec_data, GroomedData):
-            result[sec_name] = GroomedSectionMetadata(
-                type="groomed", date=sec_data.date, subsections=sec_data.subsections
-            )
+            title = _section_display_title(sec_name, sec_data.date)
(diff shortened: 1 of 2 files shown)


plugins/development-harness/backlog_core/operations.py
```diff
@@ -1977,28 +1977,35 @@ def _build_sections_from_yaml_item(item: BacklogItem) -> dict[str, SectionEntryM
...
         elif isinstance(sec_data, Section):
             entries = sec_data.entries
             entry_dicts: list[SectionEntryDict] = [
                 SectionEntryDict(id=e.id, struck=e.struck, content=e.content) for e in entries
             ]
             active_count = sum(1 for e in entries if not e.struck)
             struck_count = sum(1 for e in entries if e.struck)
-            result[sec_name] = _SectionMetadata(num_entries=active_count, num_struck=struck_count, entries=entry_dicts)
+            title = _section_display_title(sec_name)
+            result[title] = _SectionMetadata(num_entries=active_count, num_struck=struck_count, entries=entry_dicts)
```
Jamie-BitFlight: **Critical: display-title collisions silently drop section data.**

`result` is now keyed by `_section_display_title(sec_name)` instead of the raw storage key. Raw keys are guaranteed unique (`heading_to_unknown_key` always prefixes `unknown__`), but display titles are not — `_normalize_section_key` only resolves 3 built-in aliases and returns everything else verbatim, so e.g. `"Notes"` and `"notes"` are two distinct storage keys that both title-case to `"Notes"`.

Reproduced directly against this branch: grooming two sections named `"Notes"` and `"notes"` leaves only **one** in `view_item`'s result — the other's entries silently vanish (no error, no warning). Same mechanism hits a GitHub-parsed `unknown__issue_classification` colliding with the canonical `issue_classification` section (both title to `"Issue Classification"`).

The sibling function `_build_sections_metadata` (same file, ~line 2046) already solves this for its heading-keyed dict — it merges same-titled sections via `_merge_section_entries` instead of overwriting. Suggest the same treatment here, e.g.:

```python
if title in result:
    title = f"{title} ({sec_name})"  # or merge, matching _build_sections_metadata
result[title] = _SectionMetadata(...)
```

No test in this PR exercises the collision case — worth a regression test alongside the fix (two sections whose raw keys differ but display titles match).

plugins/development-harness/backlog_core/operations.py
```diff
@@ -1977,28 +1977,35 @@ def _build_sections_from_yaml_item(item: BacklogItem) -> dict[str, SectionEntryM
...
         :class:`GroomedSectionMetadata` shape (``"type": "groomed"`` discriminator).
     """
     result: dict[str, SectionEntryMetadata | GroomedSectionMetadata] = {}
     for sec_name, sec_data in item.sections.items():
         if isinstance(sec_data, GroomedData):
-            result[sec_name] = GroomedSectionMetadata(
-                type="groomed", date=sec_data.date, subsections=sec_data.subsections
-            )
+            title = _section_display_title(sec_name, sec_data.date)
+            result[title] = GroomedSectionMetadata(type="groomed", date=sec_data.date, subsections=sec_data.subsections)
```
Jamie-BitFlight, line 2005: Same collision risk as the `Section` branch below (line 2008) — a `GroomedData` section titled e.g. `"notes"` and a `Section` or another `GroomedData` titled `"Notes"` would collide here too. See the comment on line 2008 for the reproduction and suggested fix; applies symmetrically to this branch.

Jamie-BitFlight (commented):
Automated review (code-reviewer + pr-test-analyzer sub-agents). One critical finding, verified with a reproduction against this branch — see inline comments. Everything else in the diff (docstring accuracy, the GroomedData vs. Section date-arg asymmetry, test-assertion quality) checked out fine.

PR #3039, opened 2026-08-19
feat(python3-tools): build the actual astral docs corpus sync

Jamie-BitFlight:
## Summary

`sync_uv_releases.py` only ever maintained a changelog stub in `references/uv/README.md` — nothing regenerated the reference corpus itself. That was a real gap against the original astral-adoption plan's stated goal: offline docs for agents, kept fresh, the way `sync_gitlab_docs.py` already does for GitLab CI docs. The gap was documented (in the workflow comment and both archive READMEs) but never closed.

This adds the missing piece: `sync_astral_docs.py` downloads each tool's repo zip from GitHub (`astral-sh/{uv,ruff,ty}`), extracts `docs/*.md` only, grooms it (strips YAML frontmatter into a title/description pair, recursively flattens pymdownx `!!!`/`???` admonitions and `=== "tab"` blocks into plain markdown — including admonitions nested inside tabs), and atomically mirrors it into `references/<tool>/docs/`. Each archive's `README.md` gets a regenerated Documentation Index.

- Replaces the old hand-curated 5–9 file summaries (uv, ty) with a full 1:1 mirror of upstream `docs/`: 81 files (uv), 21 (ty), 15 (ruff).
- ruff had **no archive at all** before this — created one, wired into `ruff/SKILL.md`'s new `## Archive` section.
- Extended `.github/workflows/sync-astral-corpus.yml` with a second job that runs this weekly and opens a PR scoped to `references/*/docs/**` when the corpus changes, alongside the existing uv-changelog job.
- Pointed `uv`/`ty`/`ruff` `SKILL.md` Archive links at `README.md` directly instead of the bare directory — a directory link doesn't reliably make an agent open the file inside it.
- Excluded the generated `docs/` mirrors from markdownlint (same precedent as `graphify-out/**`) — it's a faithful upstream cache, not prose to rewrite for lint compliance.
- Fixed a real `.gitignore` bug found along the way: `*/.sync-*.lock` only matches one directory level, so it never actually ignored the real (4-levels-deep) lock file paths. Switched to `**/`. This exposed a stray already-committed `.sync-gitlab-docs.lock` from an unrelated past commit, now untracked.

## Test plan

- [x] Ran the script locally for all three tools (`--force`), inspected diffs
- [x] Found and fixed a real bug: nested admonitions inside tab blocks weren't groomed (single-pass parser missed them) — made grooming recursive, verified no raw `!!!`/`???`/`=== "` markers survive anywhere in the synced corpus
- [x] Found and fixed asset bloat: initial extraction pulled in images/CSS/JS/theme HTML from the zip, not just markdown — filtered to `.md` only
- [x] `uv run ruff check` / `ruff format` clean, `uv run prek run` fully green
- [x] Verified `.gitignore` fix with `git check-ignore -v` against real lock file paths (previously silent no-match, now matches)
- [x] Confirmed no other file in the repo referenced the deleted flat curated files by path before removing them


Jamie-BitFlight:
Closing without merging — after testing whether an offline docs corpus actually gets used, the value case didn't hold up. With network access available (the normal case), the agent consistently goes straight to the official astral:{uv,ty,ruff} skills / live docs.astral.sh rather than this local mirror — confirmed empirically across multiple probes. The corpus's only real use case is no-network access, which is rare for how this repo is actually used, and that value was never verified while the maintenance cost (weekly CI sync, PR review on every upstream doc change, ~33k lines of generated content) is real and ongoing. Not worth the trade.

PR #3056, opened 2026-08-20
docs(development-harness): component architecture map and contract scope binding

Jamie-BitFlight:
Adds the package-level component map this plugin did not have, and closes gaps a six-reviewer pass found in the markdown consumption contract.

Stacked on #3053 — review that first; this branch contains its commits.

## Why a component map

The plugin has eight code packages and one of them had an architecture document. Nothing described how they relate or which owns what. The practical consequence: work starting in the only documented package generalised from it, repeatedly. The `progressive_markdown` engine — which provides parsing, addressing, and pagination for markdown from any source — was invented around rather than used, and its `Navigator` and `Paginator` currently have no consumers at all.

Each package entry states role, what it owns, what it consumes, where its detail lives, and a do-not-confuse-with note wherever two components resemble each other. That last part is the load-bearing piece: the failure mode was never "could not find it," it was "found something adjacent and assumed."

## Review findings addressed

Six reviewers read existing documentation first, then the new documents, then reported contradictions. Changes made in response:

- Contract now names the operations it binds, on every transport. It previously claimed to serve the plan/task system without saying which operations that meant.
- R6 (sections and artifacts in one inventory) is marked as intended behaviour requiring a response-shape change. Three reviewers independently ranked its ambiguity highest — each had to guess whether it described current or future state.
- R8 added: paginated content is addressed by a content identity resolving against cached parsed content, so later pages are not re-collected and pages from two document versions are never stitched together. Reviewers confirmed no existing mechanism covers this.
- The plan mutation single-writer constraint from ADR-1770-1 is recorded in the map. It was omitted, and it is easy to violate silently when adding a mutation path.
- Artifact discovery ownership assigned. Three packages touch artifacts and none owned discovery.
- The dot-path address scheme and the bracket-numbered index are now explicitly distinguished, with which survives. A reviewer conflated them after reading both documents in full, which made the ambiguity mine rather than theirs.
- `backlog_core/ARCHITECTURE.md` claimed navigation ownership that belongs to the engine. Corrected to read as transitional, with a pointer to the boundary.

## Open questions

Five decisions are recorded in the contract rather than assumed: the budget default relative to the harness response cap, compact-form composition when both inventories are large, whether callers may set page size, and R8's cache scope and identity derivation.

## Sequencing note for implementation

The hand-built index structures marked for deletion are what the grooming validation gate currently reads to check section presence. That gate needs rewriting before they are removed.

Relates to #3054.

<sub>Reviews (7): Last reviewed commit: ["docs(development-harness): track the con..."](https://github.com/jamie-bitflight/claude_skills/commit/7184b5721051a987e5fbd42d545b04ca27e80608) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55006104)</sub>

### DIFF
--- a/.claude/agent-memory/dh-service-docs-maintainer/MEMORY.md
+++ b/.claude/agent-memory/dh-service-docs-maintainer/MEMORY.md
@@ -3,3 +3,8 @@
 - [DH MCP-vs-CLI Documentation Structure](dh-mcp-cli-docs.md) — canonical CLI-mapping source, which
   dh docs already pair MCP-reference sections with a dedicated CLI section, and the drift patterns
   found there (stale tool names, overstated parity, extraction-rule blind spots).
+- [Unenforced map guarantee](unenforced-map-guarantee.md) — backlog_view map mode's "under 2,000
+  tokens" claim is not enforced by disclosure_handler.py; other locations asserting the same false
+  bound; tracking issue #3059.
+- [skilllint token threshold](skilllint-token-threshold.md) — prek passing does not mean
+  skilllint's 4400-token SKILL.md ceiling still passes; re-run skilllint directly after edits.
--- a/.claude/agent-memory/dh-service-docs-maintainer/skilllint-token-threshold.md
+++ b/.claude/agent-memory/dh-service-docs-maintainer/skilllint-token-threshold.md
@@ -1,0 +1,22 @@
+---
+name: skilllint-token-threshold
+description: prek passing on a SKILL.md edit does not mean skilllint's 4400-token AS005/SK006 threshold still passes — re-run skilllint after every SKILL.md edit, not just once at the end
+metadata:
+  type: feedback
+---
+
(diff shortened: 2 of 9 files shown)


Jamie-BitFlight:
All three findings addressed.

**Component map unindexed** — fixed in `606ce76b`, which added its Required Reading entry ahead of the contract entry (it is the read-first document). That commit landed after the review ran.

**Normative range excludes R8** — fixed in `606ce76b` for `AGENTS.md` and in this push for the mechanical reference, which carried the same stale range. Good catch on the second location; I had only corrected the first.

**Engine class name incorrect** — confirmed and fixed. The class is `ProgressiveMarkdownNavigator`; the appendix named `Navigator`, which does not exist. Anyone following the appendix to migrate a duplicate implementation would have searched for a nonexistent symbol. I have also recorded in the appendix that this class and `Paginator` currently have no consumers, since that is the gap the appendix exists to describe and it was implied rather than stated.

Jamie-BitFlight:
@greptile review

Issue #3157, opened 2026-08-22
backlog_view invents sections from headings inside entry content, because operations.py keeps a second unguarded section scanner


Jamie-BitFlight:
## Story

As a **developer relying on this plugin**, I want to **backlog_groom reports success while discarding the content it was given, and a section with unbalanced entry tags absorbs the next section** so that **the tool works correctly and reliably**.

## Description

Three grooming-agent writes to #3152 returned success and persisted nothing usable. The agents' work was lost each time with no error surfaced to the caller.

Event 1 — impact-analyst wrote the Impact Radius section. `backlog_groom` returned `{"groomed_updated": true}`. The section came back holding only its scope-summary paragraph; the six required categories, Systems Inventory, and Ecosystem Completeness Checklist were absent.

Event 2 — fact-checker wrote the Fact-Check section. Same success return. The section came back containing exactly one entry whose entire content was the fragment `<div><sub>2026-08-22T14:58:39.771586Z</sub>` — no verdicts at all.

Event 3 — fact-checker retried with `replace_section` plus a reason. Same success return. The section came back containing only `<div><sub>0000-00-00T00:00:00Z</sub>\n\n<details><summary>struck: 2026-08-22T15:08:09.685995Z — Replace corrupted entry with complete fact-check verdicts</summary>`. Nine claims of verification work lost a second time.

Two contributing causes were reproduced directly and are fixed on branch `fix/3152-no-generated-story-acceptance-criteria` (commit `957194655`): `wrap_entry` double-wrapping already-wrapped content, and `_parse_entry_timestamp` raising `ValueError: year 0 is out of range` on the zero-date fallback. Neither of those accounts for what remains here.

What remains, in two parts.

Part 1 — the write path reports success when it has discarded content. Every one of the three calls returned a success payload. Nothing in the return value distinguished "wrote your content" from "wrote a truncated fragment of your content" from "wrote none of your content". A caller — human or agent — has no way to detect the loss except by reading the section back and comparing, which nothing instructs it to do. `.claude/rules/silent-failure-prevention.md` states that a function performing side effects must return a value indicating what changed, and that callers must use it. This path violates that rule at its most consequential point. Establishing whether the write itself failed, or whether the write succeeded and a later read/round-trip destroyed the content, is part of this work — the observations above cannot yet distinguish those.

Part 2 — a section whose entry tags are unbalanced absorbs the following section. Reproduced against current source:

[...]

Issue #3153, opened 2026-08-22
fix: Entry id 0000-00-00T00:00:00Z is a missing-timestamp fallback stored where date parsers read, not a marker for anything


Jamie-BitFlight:
## Story

As a **developer relying on this plugin**, I want to **entry id 0000-00-00t00:00:00z is a missing-timestamp fallback stored where date parsers read, not a marker for anything** so that **the tool works correctly and reliably**.

## Description

`backlog_core/entry_blocks.py:157` (`parse_entries`) assigns `f"{added_date}T00:00:00Z"` as the entry id for any section body that contains no `<sub>` entry wrapper. `added_date` itself falls back to the literal string `"0000-00-00"` in several places: `parsing.py:368` (`_fm_str(fm, meta, "added") or "0000-00-00"`), the default parameter on `parsing.py:1188` `parse_md_body_sections`, `entry_blocks.py:134` and `:266`, and `operations.py:1021`, `:1082`, `:1166`, `:1213`. `entry_blocks.rewrite_section` writes the same value through `wrap_entry_with_timestamp` when it wraps legacy unwrapped content.

The composed result is `0000-00-00T00:00:00Z`. That is not a valid ISO 8601 timestamp — no year zero, no month zero, no day zero. It occupies the same field that every authored entry uses for a genuine `now_iso()` timestamp, so one field holds two kinds of value.

Observed on #3152: every section of that item — including the authored `Description` and the authored first `Context` entry — carries `0000-00-00T00:00:00Z`, while the item's own frontmatter records `added: 2026-08-22`. So the real added date exists and was not threaded through to the entries that were parsed from the freshly created issue body.

Two distinct problems appear to be tangled here, and grooming should separate them:

1. The fallback fires when a real date is available. #3152 has `added: 2026-08-22` in its metadata yet its entries carry the zero date. Whichever parse path built those entries did not receive the item's `added` value, or read it from a location that was empty at parse time.

2. The fallback value itself is unparseable. Even where no added date exists, emitting a string that no date parser should accept puts unparseable data in a field that is read as a date. `backlog_view` accepts a `since` parameter documented as "ISO date/datetime. Only entries at or after this timestamp are included," and `parse_entries` filters on `_parse_entry_timestamp(e.id) >= since_dt`. What `_parse_entry_timestamp` does with the zero value — raise, coerce, or sort first — has not been tested and should be established as part of grooming rather than assumed.

[...]

PR #3160, opened 2026-08-22
fix(backlog-core): stop generating unauthored sections at item creation, and stop the write path losing content

### LINKED ISSUE #3152: fix: Created items ship with a generated Story and Acceptance Criteria that no author wrote and that can invert the item's meaning
## Story

As a **developer relying on this plugin**, I want to **created items ship with a generated story and acceptance criteria that no author wrote and that can invert the item's meaning** so that **the tool works correctly and reliably**.

## Description

Every backlog item created through `backlog_add` / `backlog add` ships with a Story and an Acceptance Criteria section that no author wrote. `backlog_core/parsing.py` template-fills them at creation time from the item title.

Observed on #3151, created 2026-08-22. The generated Story:

    As a **maintainer of the codebase**, I want to **file-based language for sam tasks
    persists because the context-file key and agents.md required reading still teach it**
    so that **the code is cleaner and more maintainable**.

The title was lowercased and dropped into an `I want to {goal}` slot. The result is ungrammatical, and it states the opposite of the item's intent — it reads as wanting the file-based language to persist, when the item exists to remove it. The benefit clause is a fixed string unrelated to the item.

The generated Acceptance Criteria:

Both assert nothing checkable. Neither describes this item or any other.

Both entries carry the entry id `0000-00-00T00:00:00Z`, a sentinel distinguishing them from authored entries. So the system records that they are generated, but a consumer reading sections by name — `backlog_view(sections=["Acceptance Criteria"])`, a groomer's validation gate, an agent deciding whether an item is ready — sees a populated section with plausible-looking content and no signal that nobody wrote it.

Impact observed: a section that exists and is populated reads as authored. On #3151 the placeholder criteria sat alongside genuine observations, and the fabricated Story asserted the inverse of the stated problem. An item whose Acceptance Criteria section is non-empty may satisfy a completeness check that was meant to confirm someone had defined criteria.

Ordering question this raises, for grooming to settle: acceptance criteria and a user story are grooming outputs in the agile sense — they are established during refinement, with the information grooming gathers, not at intake when only the problem is known. The create flow currently produces them before that information exists. An alternative shape is that creation supplies the template unfilled, or supplies nothing, and the grooming workflow writes both — `groom/finalize.md` already writes sections through `backlog_groom(..., mark_groomed=True)` and already runs a required-sections validation gate.

Related observation, possibly a separate defect: on #3151 the item metadata carries `groomed: "2026-08-22"` while its label is `status:needs-grooming`. `backlog_groom` calls stamped a groom date without `mark_groomed=True`, so the date field and the status label disagree about whether the item has been groomed.

## Acceptance Criteria

## Context

- **Source**: Session observation — 2026-08-22, observed on #3151 at creation
- **Priority**: P1
- **Added**: 2026-08-22

Jamie-BitFlight:
Closes #3152. Contributes to #3153. Partially addresses #3157.

### 1. Creation no longer fabricates Story and Acceptance Criteria (#3152)

`build_issue_body()` template-filled two sections from the item title. The Story slotted `title.lower()` into an `I want to {goal}` clause — which **inverts the meaning** of any item titled after the thing to be removed — and paired it with a fixed benefit string. The Acceptance Criteria section was two constant, uncheckable placeholders.

Both are grooming outputs. Worse, `groom/finalize.md` gates on a non-empty Acceptance Criteria section, so a section emitted at creation satisfied that gate with nobody having written a criterion.

Creation now emits only sections whose content the caller supplied. `ROLE_MAP` and `BENEFIT_MAP` had no other consumers and are deleted.

Reproduction, using the #3151 title that triggered the report — before:

```
As a **maintainer of the codebase**, I want to **file-based language for sam tasks
persists because the context-file key and agents.md required reading still teach it**
so that **the code is cleaner and more maintainable**.
```

After: the body contains Description and Context only.

### 2. `wrap_entry` no longer nests wrappers and loses content

`backlog_view` renders entries with their `<div><sub>...</sub>` wrapper visible. An agent that reads a section to match house style and writes back in that style submits pre-wrapped content, which was then wrapped again. The nested form does not survive a body round-trip: the wrapper leaks into the entry's own content and the section splitter emits a stray `</div>` entry with an empty ID.

Observed live three times during a single grooming run of #3152 — each `backlog_groom` call returned success while discarding the agent's work.

### 3. The `since` filter no longer crashes on the zero-date fallback

`parse_entries` assigns `f"{added_date}T00:00:00Z"` for any section with no `<sub>` wrapper, and `added_date` itself falls back to `"0000-00-00"`. `datetime.fromisoformat` rejects that with `ValueError: year 0 is out of range`, so `backlog_view(since=...)` crashed on any item holding legacy unwrapped content. #3152 recorded this as untested; it is tested now.

The prefix means "timestamp unknown", not "year zero" — it maps to `datetime.min` (UTC), sorting before every real timestamp and excluded by any real cutoff. Scoped to that sentinel: genuinely malformed IDs, including the empty ID the nesting corruption produced, still raise.

### 4. Documentation corrected

`skills/backlog/references/item-schema.md` and `skills/backlog/README.md` — both canonical — named `create-backlog-item` as the Acceptance Criteria writer. Now `groom-backlog-item`.

## Verification

Claims carrying the fix's safety argument were verified repo-wide, independently, twice:

| Claim | Verdict | Evidence |
|---|---|---|
| `build_issue_body` has one non-test caller, create path only | VERIFIED | `gh_client.py:46` import, `:1277` call, inside the sole issue-creation function |
| `Story` in neither required nor optional list in `finalize.md` | VERIFIED | `finalize.md:139-150` enumerate all validated sections |
| No title-derived Story generator remains | VERIFIED | only `## Story` hit in `parsing.py` is a docstring at `:754` |
| A second producer reintroduces the bug | REJECTED | `group-items-to-milestone` instructs an agent to author a story-format body — not title derivation, cannot invert meaning |

Tests: 10 new regression cases in `backlog_core/tests/test_entry_wrapper_nesting.py`, plus `build_issue_body` tests inverted from `contains_` to `omits_` and a new guard that no body content is ever derived from the title. Full suites: **3385 passed, 39 skipped, 5 xfailed**.

## Not fixed here

#3157 remains open for the two parts this PR does not close: the write path still reports `groomed_updated: true` when it has discarded content, and a section with unbalanced entry tags still absorbs the following section (`ENTRY_RE` has no boundary awareness). Both need a parser change across several modules rather than a local patch.

## Note for reviewers

`models.py` conflicts are possible with #3145 and #3147, which touch the same file in a different region.

https://claude.ai/code/session_01WaKHjD7ZWU8JFRzTzUJGFS

### DIFF
--- a/plugins/development-harness/.claude-plugin/plugin.json
+++ b/plugins/development-harness/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dh",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
-  "version": "10.0.2",
+  "version": "10.0.3",
   "author": {
     "name": "Jamie Nelson",
     "url": "https://github.com/bitflight-devops"
--- a/plugins/development-harness/.codex-plugin/plugin.json
+++ b/plugins/development-harness/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "dh",
-  "version": "9.0.2",
+  "version": "9.0.3",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
   "author": {
     "name": "Jamie Nelson",
--- a/plugins/development-harness/.cursor-plugin/plugin.json
+++ b/plugins/development-harness/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dh",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
-  "version": "6.0.2",
+  "version": "6.0.3",
   "author": {
     "name": "Jamie Nelson",
     "url": "https://github.com/bitflight-devops"
--- a/plugins/development-harness/backlog_core/ARCHITECTURE.md
+++ b/plugins/development-harness/backlog_core/ARCHITECTURE.md
@@ -215,7 +215,7 @@ Functions that previously raised `typer.Exit(1)` must instead raise one of:
 
 **Functions/data extracted from backlog.py** (line references are approximate):
 
-- Constants: `BACKLOG_DIR`, `DEFAULT_REPO`, `SECTION_RE`, `SKIP_STATUS`, `GITHUB_ISSUE_URL_RE`, `GITHUB_ISSUE_TITLE_TRUNCATE`, `MIN_FRONTMATTER_PARTS`, `TYPE_TO_LABEL`, `ROLE_MAP`, `BENEFIT_MAP`, `FUZZY_DUPLICATE_THRESHOLD`, `_COMMIT_PREFIX_RE`, `_FIELD_TO_INDEX`
+- Constants: `BACKLOG_DIR`, `DEFAULT_REPO`, `SECTION_RE`, `SKIP_STATUS`, `GITHUB_ISSUE_URL_RE`, `GITHUB_ISSUE_TITLE_TRUNCATE`, `MIN_FRONTMATTER_PARTS`, `TYPE_TO_LABEL`, `FUZZY_DUPLICATE_THRESHOLD`, `_COMMIT_PREFIX_RE`, `_FIELD_TO_INDEX`
 - Add new: `PRIORITY_SECTIONS` dict mapping priority strings to section headings (from the `add` command)
 - Exception classes: `BacklogError`, `ItemNotFoundError`, `DuplicateItemError`, `AmbiguousSelectorError`, `GitHubUnavailableError`, `ValidationError`
 - Pydantic models: `Entry`, `Section`, `GroomedData`, `BacklogItem`, `Output`, `IssueStatus`, `PullRequestRef`, `ViewItemResult`, `IssueLocalFields`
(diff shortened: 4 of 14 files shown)


Jamie-BitFlight:
## Correction to this PR's description

The description attributes the three observed grooming content-loss events to `wrap_entry` nesting. **That attribution is wrong** and I am correcting it here rather than silently editing the description.

### What actually happened

The damage on #3152 is section explosion, not truncation. The item now carries 46 top-level sections. Sections named `Code — Producers (write the changed interface)`, `Code — Consumers (read the changed interface)`, `Documentation (will become stale)`, `Systems Inventory`, and `Ecosystem Completeness Checklist` are the impact-analyst agent's `###` subheadings promoted to top level. Roughly 28 sections named `Claim A:` through `Claim I:` are the fact-checker's subheadings, repeated across its three attempts.

Nothing was truncated. Content was split on its own markdown headings, and what appeared to "survive" in each section was simply the text preceding the first `###`.

### Why the original attribution was wrong

I built the reproduction to match my hypothesis — nested `<div>` wrappers with no headings — rather than using the agents' actual submitted content, which contained both nesting and `###` headings. The reproduction passed and proved nothing about the real failure.

### What I cannot reproduce against current source

No combination produces the explosion:

- wrapped entry content containing `###` headings, through `extract_sections` — no split
- the same with `##` headings — no split
- pre-wrapped (nested) content containing `###` headings — no split
- `render_issue_body` → `parse_issue_body` round trip with `###` inside an entry — clean

`_EntryDivBlock` in `_split_body_h2` holds in all of them. It landed in `bb8ee0c0a` ("unify section-key normalization between local writes and GitHub-parsed sections (#2964)") on 2026-08-18 and is on main. `extract_sections`'s docstring states that the marko splitter replaced a "naive line-based `## ` scan" which misidentified heading-shaped lines inside entry content as section boundaries — exactly the observed behaviour. The leading hypothesis is now that the live failures came from the MCP server running a cached build predating that commit. This is under investigation and is **not** yet established.

### What is unaffected by this correction

The three code changes in this PR stand on their own evidence and are independently reproduced and tested:

1. **#3152 — creation no longer fabricates Story and Acceptance Criteria.** Unrelated to the above. Verified repo-wide, twice, independently.
2. **`wrap_entry` no longer nests wrappers.** The nesting defect is real and reproduced — it produces a leaked wrapper in entry content and an orphan entry with an empty ID. It was simply not the cause of the three grooming failures.
3. **The `since` filter no longer raises `ValueError: year 0 is out of range`.** Real, reproduced, unrelated to the above.

No change to the diff is required. The claim being corrected is in the prose, not the code.

### Consequence for #3157

#3157 is filed on the premise that a parser boundary bug is live in source. That premise may be false. Its second part — the write path returning `groomed_updated: true` regardless of what actually persisted — remains real and worth fixing regardless of what caused the explosion. #3157 will be re-scoped once the investigation reports.

Issue #3113, opened 2026-08-21
docs: task-worker adopts orchestrator role from borrowed skill prose instead of its assigned worker role


Jamie-BitFlight:
## Story

As a **developer reading the documentation**, I want to **task-worker adopts orchestrator role from borrowed skill prose instead of its assigned worker role** so that **documentation is accurate and trustworthy**.

## Description

During #3060's implementation, `work-3060` delegated the entire `dh:implement-feature` skill invocation to a `dh:task-worker` subagent instead of running it inline itself. `implement-feature/SKILL.md` is written in first person for "the orchestrator" (query ready tasks, TeamCreate, dispatch task-workers). The receiving task-worker complied literally with that borrowed voice, adopting a coordinator role it was never assigned, and cascaded into TeamCreate plus further Agent dispatches (~$100 unplanned spend).

Root cause is textual, not permission-based. `task-worker.md`'s Identity section ("you become whatever the task requires by loading the right skills... your job is to do the work — not to ask the manager how to do it") provides no resistance to executing orchestrator-voiced skill content it's handed, and actively discourages the pushback that should fire on a role conflict. `implement-feature/SKILL.md`'s use of "the orchestrator" to mean "whoever happens to be executing this skill" is ambiguous against the plugin's actual, verified reality: exactly one agent per session — the interactive Claude Code session — is ever the orchestrator (confirmed by direct comparison of system prompts: the interactive session identifies itself as "an interactive agent"; a dispatched subagent's system prompt carries no such framing at all). Every dispatched agent, however many levels deep, and regardless of whether it itself dispatches further agents, acts on behalf of its own dispatcher — never on behalf of the human directly.

Sub-dispatch by a subagent is legitimate and valuable when explicitly assigned by its own dispatcher (a "manager" role, distinct from "orchestrator") — a subagent dispatching further subagents as planned decomposition is a recently-added platform capability, not inherently wrong. The defect is specifically role self-assignment — inferring a coordinating role from a loaded skill's own authorial voice rather than from an explicit assignment. This also predates a recent (~8 week) platform capability change (subagents gained the ability to dispatch subagents) that this workflow's prose was never revisited against, which is why the ambiguity existed unnoticed: previously nothing but the top-level session could ever have been in a position to misread "the orchestrator" as self-referential.

[...]

PR #3173, opened 2026-08-23
docs(development-harness): defer the dispatch target to dh:dispatch-contract

Jamie-BitFlight:
Part of #3113

## What

`role-resolution-protocol.md` carried a universal under `**Dispatch routing rule:**`: regardless of
what a role resolves to, the orchestrator always dispatches `subagent_type="dh:task-worker"`.

Replaced with: record the resolved agent name in the task's `agent:` field, then choose the dispatch
target with the decision in `dh:dispatch-contract`.

## Why the universal is the wrong half

Verified against source on `main`, not taken from the audit report:

- `main` contradicts its own universal in four places — `skills/forensic-review/SKILL.md` dispatches
  `dh:code-reviewer`, `skills/implement-feature/SKILL.md` dispatches `dh:contract-verification`,
  `skills/groom-backlog-item/references/groomer-agent.md` and
  `work-backlog-item/references/workflows/groom/swarm.md` dispatch `dh:backlog-item-groomer`,
  `skills/backlog-tools-administrator/SKILL.md` dispatches
  `python3-development:python-cli-architect` and `plugin-creator:ai-doc-optimizer`. Direct specialist
  dispatch predates the dispatch-contract skill.
- Scoping the rule by originating plugin would be wrong. `plugins/python3-development/skills/python3-development/references/language-manifest.md`
  declares `- code-reviewer: @dh:code-reviewer`, and `language-manifest-schema.md` validates
  `@plugin:agent` syntax only — never that the plugin differs from `dh`. A project-local
  `.claude/language-manifest.md` override can name any installed agent. The discriminator is the
  resolved agent's declared capability, which is what `dh:dispatch-contract` already decides.

The adjacent `**Resolution rules:**` block is untouched. It maps abstract roles to agent names, a
different operation, and it remains correct.

## Ordering dependency

`dh:dispatch-contract` does not exist on `main` yet — it ships in #3134, which already carries this
exact one-line replacement. Merge #3134 first, or merge this and accept the pointer resolving only
once #3134 lands. If #3134 merges first this PR becomes a no-op and can be closed.

## Out of scope, found while verifying

`skills/add-new-feature/SKILL.md` states "Phase 3 always dispatches `subagent_type=\"dh:task-worker\"`".
That is a phase-specific decision rather than the universal fixed here, and #3134 does not touch that
file either. Worth a follow-up.

### DIFF
--- a/plugins/development-harness/.claude-plugin/plugin.json
+++ b/plugins/development-harness/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dh",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
-  "version": "10.0.1",
+  "version": "10.0.2",
   "author": {
     "name": "Jamie Nelson",
     "url": "https://github.com/bitflight-devops"
--- a/plugins/development-harness/.codex-plugin/plugin.json
+++ b/plugins/development-harness/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "dh",
-  "version": "9.0.1",
+  "version": "9.0.2",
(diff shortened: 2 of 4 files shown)


Jamie-BitFlight:
Superseded by #3134, which landed the identical replacement for the dispatch routing rule and ships the `dh:dispatch-contract` skill this change points at. Verified on main after the merge: the paragraph now reads "Record the resolved agent name in the task's `agent:` field, then choose the dispatch target with the decision in `dh:dispatch-contract`." No content from this branch is lost.

PR #3338, opened 2026-08-30
fix(receiving-pr-reviews): port seven vendored-skill fixes back from skilllint

Jamie-BitFlight:
`.agents/skills/` is vendored into [`bitflight-devops/skilllint`](https://github.com/bitflight-devops/skilllint). Codex review of the vendored copy found seven defects, fixed there in bitflight-devops/skilllint#119 and bitflight-devops/skilllint#134. This ports them to their canonical home before the next sync reverts them. Tracking issue: bitflight-devops/skilllint#135.

### `.agents/skills/receiving-pr-reviews/scripts/pr_review_threads.py`

1. **Deleted `_MIN_POLL_BUDGET_SECONDS = 5.0`** — an unsourced constant that could both skip the final poll *and* exit 1 on an ordinary run. `deadline` is now the only cutoff; the loop polls while a full `--interval-seconds` still fits and stops once less than that remains.
2. **Deleted `_GH_TIMEOUT_SECONDS = 30`** — also unsourced, and it capped an entire sequential `gh api --paginate` run rather than a single request. Replaced by a `--gh-timeout-seconds` option on both `fetch` and `watch`, unbounded by default.
3. **Strict Pydantic ingress models** — every model that ingests a raw GitHub GraphQL response now inherits `_GitHubResponseModel` (`ConfigDict(strict=True)`), so a producer-shape mismatch from `gh` fails at the boundary instead of being coerced into apparently valid review state. Models this script builds itself from already-validated values are not ingress and do not inherit it.
4. **Unbounded baseline fetch, plus `min` constraints** — `watch`'s mandatory baseline is no longer deadline-bounded, so `--timeout-seconds 0` returns the documented immediate snapshot instead of raising `TimeoutExpired`. `--interval-seconds` gains `min=1`, `--timeout-seconds` `min=0`, `--gh-timeout-seconds` `min=0`.
5. **Comment-identity activity key** — `_thread_activity_key` now returns `(comments_total, ((databaseId, body), ...))`. Diffing on `databaseId` and `comments_total` alone missed an *edit* to an existing comment, which moves neither, so `watch` reported `timed_out: true` while revised feedback sat unread. Three change kinds must register and no single component catches all three: a reply adds an id, an edit changes only the body, and a reply past the query's `comments(first: 100)` page moves only the untruncated `comments_total`.
6. **Split exception handlers** — `TimeoutExpired` and `CalledProcessError` are no longer handled together. Only a timeout is explained by the deliberately shrinking budget, so at or past `deadline` it is an honest end-of-window; a non-zero `gh` exit (auth, rate limit, API, GraphQL) is a failed poll whatever the clock says.

### `.agents/skills/receiving-pr-reviews/SKILL.md`

7. **Two corrected gotchas** — the near-deadline poll gotcha now describes the actual behaviour (last observed state can be up to one interval stale; `timed_out: true` means "issue another call", never "the window through `deadline` was checked"), and the activity-detection gotcha now says a thread counts when its *comments* changed, not merely when its `id` is new.

### `.claude/skills/rebase/SKILL.md`

8. **`git rebase <target> <branch>` positional form plus worktree-ownership guidance** — the positional form is `[<upstream> [<branch>]]`, so omitting `<branch>` rebased whatever was checked out rather than the branch the skill's own Step 1-4 plan named. Naming `<branch>` makes git check it out first, which fails when another worktree holds it, so the skill now says to run the rebase from the worktree that owns the branch rather than detaching or switching someone else's.

(Seven fixes, eight bullets: the two SKILL.md gotchas are counted as one fix in the issue.)

## Already upstream / skipped

None of the seven were present on `main`. Nothing was skipped.

`DEFAULT_OWNER`/`DEFAULT_REPO` and the corresponding SKILL.md repo references stay `Jamie-BitFlight`/`claude_skills` — skilllint's vendored copy retargets them at itself, and that retargeting is deliberately *not* part of this port.

## Two extra changes, not among the seven

This repository lints `.agents/skills/` with ruff; skilllint excludes it as vendored code. Two ruff findings surfaced only here:

- `_run_gh`'s docstring had `timeout:` indented as a continuation of `args:`, so the parameter was undocumented (`D417`). Dedented.
- Two `_gh_timeout_budget` assertions used `==` on floats (`FURB` float-equality). Now `pytest.approx`.

Both are defects in the skilllint copy too and are worth back-porting there.

## Overlap with in-flight work

- `.claude/worktrees/fix-watch-timeout-starvation` (`worktree-fix-watch-timeout-starvation`): **no overlap** — its `receiving-pr-reviews` content is byte-identical to `main`, so it is already merged. Untouched by this PR.
- `fix/pr-review-watch-stateless-signals` (pushed, unmerged): **real overlap, needs a human decision.** That branch splits the script into `pr_review_gh.py` / `pr_review_models.py` and makes `watch` stateless — triggering on absolute counts via `has_outstanding_work()` rather than a per-call baseline diff. That is a *different* solution to the same concern as fix 5, and it deletes the baseline this PR fixes. It also still carries `_MIN_POLL_BUDGET_SECONDS = 5.0` (fix 1) and the combined exception handler (fix 6). Whichever lands second will need a real merge, not a rebase.
- `docs/pr-review-reply-destination` and `fix/no-derived-data-sweep`: already merged into `main`, no overlap.

## Verification

- `uv run python -c "import ast, pathlib; ast.parse(...)"` on both scripts: parses.
- `uv run pytest .agents/skills/receiving-pr-reviews/scripts/test_pr_review_threads.py -q`: **17 passed**.
- `uv run ruff check --no-fix` + `uv run ruff format`: clean.
- `uv run ty check`: `All checks passed!`
- `prek run --files <the four changed files>`: every applicable hook passed, including `markdownlint-cli2`, `skilllint check --fix`, `ty`, and the conventional-commit gate.


Jamie-BitFlight:
Closing in favour of #3340 and #3341, which replace this branch. Not abandoned — split, because #3337 landed underneath it.

**What #3337 changed.** It split `pr_review_threads.py` into `pr_review_threads.py` / `pr_review_gh.py` / `pr_review_models.py` and made `watch` stateless, triggering on absolute counts via `has_outstanding_work()` instead of a per-call baseline diff. That is a design change to the same concern one of these seven fixes addressed, so this stopped being a rebase and became a merge between two solutions — which is why it was left conflicting rather than force-resolved.

**#3337 superseded exactly one fix, and it is the better design.** The comment-identity activity key here keyed the baseline diff on `(comments_total, ((databaseId, body), ...))` so an edited comment would register. Deleting the baseline removes that bug class rather than patching the key: `unresolved_count > 0` is true whether a comment was added, edited, or sits past the 100-comment page, so `watch` never sleeps through any of them. Dropped, not merged — porting it would have been a regression.

**The other five were still absent from `main`, and three had got worse:**

- `_MIN_POLL_BUDGET_SECONDS` was still there and had been *raised* 5.0 → 20.0 rather than removed. Still unsourced.
- `_GH_TIMEOUT_SECONDS = 30` was still there, and now the default on **seven** fetch functions instead of two — one cap over strictly more sequential work.
- The ingress surface **grew** by five lax models (`Reaction`, `IssueComment`, `GitHubCommitDate`, `HeadCommitNode`, `ForcePushEvent`), none strict.
- The first fetch was still deadline-bounded, so `--timeout-seconds 0` still starved the documented immediate snapshot.
- The exception handlers were still combined — which, with the shrinking budget, is why `main` can exit 1 on an ordinary run today.

So the work was re-implemented against the new layout rather than cherry-picked.

**The replacements:**

- **#3340** — `.claude/skills/rebase/SKILL.md` alone. #3337 never touched it, it merges clean, and it is the fix that bit a live session directly. No reason for it to wait.
- **#3341** — the five remaining fixes, re-implemented against the split layout. `_MIN_POLL_BUDGET_SECONDS` deletion and the handler split land in one commit because neither half is safe alone.

One correction to what this PR's body claimed: the `D417` `_run_gh` docstring defect (filed downstream as bitflight-devops/skilllint#143) **does not exist on `main`** — it was introduced by skilllint's own docstring rewrite and is downstream-only. `main` was correctly indented all along; #3341 keeps it that way.

Branch kept for now; safe to delete once #3340 and #3341 are merged.

Issue #134, opened 2026-02-21
Validate and verify orchestrator-discipline plugin hooks and processes


Jamie-BitFlight:
**Status**: DONE — commits 49e0ae5, ea3e737 on branch claude/bulk-backlog-grooming-LIQDi

T1-T4 complete. Plugin passes `claude plugin validate`. Hook directory detection added. `user-invocable: true` added to SKILL.md. All 5 hook behavior tests pass.

**Plan**: plan/tasks-4-validate-orchestrator-discipline.md
**Completed**: 2026-02-21

---
*Imported from `.claude/BACKLOG.md`*

PR #3341, opened 2026-08-30
fix(receiving-pr-reviews): source watch's timeouts and strict-validate gh input

Jamie-BitFlight:
Five of the seven defects Codex found in this skill's vendored copy in [`bitflight-devops/skilllint`](https://github.com/bitflight-devops/skilllint) (fixed there in bitflight-devops/skilllint#119 and #134), **re-implemented** against the layout #3337 introduced. Replaces the `receiving-pr-reviews` half of #3338, which #3337 made unrebaseable. Tracking issue: bitflight-devops/skilllint#135.

This is not a port. #3337 split the script into `pr_review_threads.py` / `pr_review_gh.py` / `pr_review_models.py` and deleted the per-call baseline these fixes were written against, so the timeout machinery moved files and one fix stopped applying entirely.

## `_MIN_POLL_BUDGET_SECONDS` and the handler split — one commit, deliberately

These are one mechanism, and `main` today **exits 1 on an ordinary run** because only half of it exists:

1. `gh_timeout_budget` deliberately shrinks each call to the time left before `deadline`.
2. The last poll of a window is therefore *expected* to be cut short — it raises `TimeoutExpired`.
3. `except (subprocess.TimeoutExpired, subprocess.CalledProcessError)` cannot tell that from a genuine `gh` failure, so it sets `last_poll_ok = False`.
4. `watch` exits 1 with nothing on stdout.

`_MIN_POLL_BUDGET_SECONDS` was the mitigation: reserve enough time that the final poll usually succeeds. It was `5.0` when `build_fetch_result` made two `gh` calls, was raised to `20.0` when it grew to seven, and its own comment concedes it is *"still an unmeasured heuristic, not a proven-sufficient margin — the exception handler around the poll in `watch` is the real backstop, not this guard alone"*. That is the argument for fixing the handler rather than raising the number a third time.

Deleting the constant without splitting the handler makes the exit-1 more frequent. Splitting the handler without deleting the constant leaves an unsourced 20-second hole in every 270-second window. Neither ships alone, so both are in one commit.

**After:** `deadline` is the only cutoff. The loop polls while a full `--interval-seconds` still fits and stops once it does not. A `TimeoutExpired` at or past `deadline` is the window ending — the same honest "no time left to check again" the loop already reports when it stops before polling. A `TimeoutExpired` with time still on the clock is a real stall. A `CalledProcessError` (auth, rate limit, API, GraphQL) is a failed poll whatever the clock says, because the deadline can neither cause nor excuse it.

The final sub-interval stretch is now left unpolled by design. That is not a regression — the next `watch` call's own first fetch covers it, which is exactly what step 7's documented loop of back-to-back calls is for.

## `_GH_TIMEOUT_SECONDS = 30` → `--gh-timeout-seconds`

Unsourced, and it caps an entire sequential `gh api --paginate` run — every page, inside one subprocess — not a single request. It was already unsourced when it covered two calls; `build_fetch_result` now makes **seven** per snapshot, so one 30-second cap covers strictly more work than the number was (not) chosen for. Threaded through all seven fetch helpers, unbounded by default, exposed on both `fetch` and `watch`. Inside `watch`, each poll stays bounded by `--timeout-seconds`.

## Strict Pydantic ingress

`_GitHubResponseModel` carries `ConfigDict(strict=True)`; every model validated from a raw `gh` response inherits it — including the five #3337 added (`Reaction`, `IssueComment`, `GitHubCommitDate`, `HeadCommitNode`, `ForcePushEvent`). Lax mode turned `"totalCount": "0"` into `0` and `"isResolved": "false"` into `True` without a word. `FetchResult`, `WatchResult` and `UnresolvedThread` are assembled from already-validated values, so they are output shapes and deliberately stay non-strict.

One wrinkle worth flagging, since it is a real constraint rather than an oversight: these models are validated from `json.loads` output — Pydantic's **Python** mode — where a strict `datetime` accepts only a `datetime` instance and would reject the ISO-8601 string GitHub actually sends. `GitHubCommitDate.committedDate` is the sharpest case: coercing a `datetime` out of `gh` output in lax mode is exactly the producer-shape silence strict mode exists to remove, but the string itself is the correct wire shape, not a mismatch. So timestamp fields take `GitHubTimestamp = Annotated[datetime, Field(strict=False)]` — strictness relaxed on exactly those fields and nothing else. The ceiling: a lax `datetime` still accepts a Unix epoch number. `test_ingress_models_still_parse_github_iso_timestamps` pins both halves.

## Unbounded first fetch, plus `min` constraints

`watch`'s first fetch was `deadline`-bounded, so `--timeout-seconds 0` — the documented immediate snapshot — floored that mandatory call to nothing and raised `TimeoutExpired`. It now takes `--gh-timeout-seconds`; only the polls race the deadline. `--interval-seconds` gains `min=1` (0 busy-loops `gh` until the timeout), `--timeout-seconds` `min=0`, `--gh-timeout-seconds` `min=0`.

## Dropped: the comment-identity activity key

The sixth skilllint fix is **not** ported. It keyed the baseline diff on `(comments_total, ((databaseId, body), ...))` so an edited comment would register. #3337 deleted the baseline entirely and triggers on absolute counts via `has_outstanding_work()` — `unresolved_count > 0` is true whether a comment was added, edited, or sits past the 100-comment page, so `watch` never sleeps through any of them. That removes the bug class instead of patching the key, and it is the better design. Nothing to merge; the port would have been a regression.

## SKILL.md

Two gotchas corrected: the final-poll reservation no longer exists, and a poll cut short by the deadline is no longer reported as a failure. A third bullet documents `--gh-timeout-seconds`. The activity-detection gotcha needed nothing — #3337 already made it accurate.

## Correction to the record

While porting I reported a `D417` defect in `_run_gh`'s docstring (`timeout:` indented one level too deep, so the parameter reads as undocumented) and filed it downstream as bitflight-devops/skilllint#143. **That defect does not exist on `main`** — it was introduced by skilllint's own rewrite of that docstring and is downstream-only. `main`'s copy was correctly indented, and this PR's rewrite keeps it that way. Verified: ruff is clean on untouched `main` for these paths. skilllint#143 stays valid downstream; nothing to fix here.

## Verification

Baseline on untouched `main` for these paths: **42 passed**, ruff clean, ty clean.

```
$ uv run pytest .agents/skills/receiving-pr-reviews/scripts/ -q
53 passed in 2.51s

$ uv run ruff check --no-fix .agents/skills/receiving-pr-reviews/scripts/
All checks passed!

$ uv run ruff format --check .agents/skills/receiving-pr-reviews/scripts/
4 files already formatted

$ uv run ty check .agents/skills/receiving-pr-reviews/scripts/
All checks passed!
```

`prek run --files <the five changed files>` — every applicable hook passed, including `markdownlint-cli2`, `ty check`, and the conventional-commit gate.

42 → 53. New coverage: the one-interval cutoff, deadline-truncated vs genuine poll failure, non-zero `gh` exit at the deadline, both `min` constraints, the unbounded first fetch, `gh_timeout_budget`'s two modes, strict rejection of coerced producer shapes, ISO-8601 timestamps still parsing, and the internal models staying non-strict.


Jamie-BitFlight:
Codex reviewed the superseded #3338, which carried the same `run_gh` change, and found a gap this PR shared. Addressed in 73433ed:

> **Restore a default timeout for gh subprocesses** — When `gh` stalls because of a network, authentication, or API issue, changing `_run_gh`'s default to `None` makes `fetch`, the `watch` baseline, `reply`, and `resolve` wait indefinitely unless the caller explicitly supplies a timeout — and the latter two commands expose no timeout option at all.

**Accepted, in part.** `reply` and `resolve` genuinely had no way to be bounded — `--gh-timeout-seconds` reached only `fetch` and `watch`. Both now take it, so every `gh`-backed command in the script can be bounded by its caller. Covered by a parametrized test over all three.

**Declined, in part:** restoring a bounded *default*. That is the finding this PR exists to fix. 30 seconds was never sourced, and one such cap has to cover a whole sequential `gh api --paginate` run — every page, inside one subprocess — on an arbitrarily large PR over an arbitrarily slow link. `build_fetch_result` now makes seven of those per snapshot rather than the two the number was (not) chosen for. Replacing it with a different unsourced number would repeat the defect at a new magnitude.

The unattended-hang concern is real but is already answered for the path where it matters: inside `watch`, every poll is bounded by `--timeout-seconds`, which the caller chose, so an unattended watch loop cannot hang regardless of `--gh-timeout-seconds`. What is left unbounded by default is a single `fetch`, `reply`, or `resolve` — foreground, one-shot calls where a caller that wants a ceiling can now pass one.

Happy to add a default if a sourced number exists; I could not find one in this repo or in `gh`'s own documentation.

Jamie-BitFlight:

PR #3343, opened 2026-08-30
feat(receiving-pr-reviews): report whether the PR can be reviewed at all

Jamie-BitFlight:
> one more feature to add to the slim output of the pr checking script, check if it is still in mergable non-conflict state and non draft. and say those things in its output if either is not true. because reviews wont happen when its not mergable or not set to ready for review.

A draft PR gets no reviewers requested. A conflicting one gets no review runs. In both cases `fetch` reported `unresolved_count: 0` — which reads as **"nothing to do"** when the truth is **"nothing can happen until the PR itself is fixed"**. That is the same misleading-empty-result trap the existing `reviews_count` / `threads_count` / `unresolved_count` gotcha already warns about; this closes the remaining hole in it.

## Output

`fetch` and `watch` both carry a `reviewability` block. Verified against live PRs — this is real output, not an example:

```jsonc
// PR #3341 — ready, conflict-free
"reviewability": {"is_draft": false, "mergeable": "MERGEABLE", "merge_state_status": "CLEAN", "blockers": []}

// PR #3338 — genuinely conflicting
"reviewability": {
  "is_draft": false,
  "mergeable": "CONFLICTING",
  "merge_state_status": "DIRTY",
  "blockers": ["conflicting: reviews will not run until the merge conflicts are resolved"]
}
```

Empty `blockers` means reviews can proceed. Each entry names the **consequence** rather than the state name, because the reader needs to know what will not happen, not what the API said.

`watch` carries it for the same reason it matters at all: blocking 270 seconds waiting for reviews that cannot arrive is pure waste.

## `UNKNOWN` is deliberately not a blocker

GitHub computes mergeability in a background job and returns `UNKNOWN` while it runs — which is exactly the moment just after a push, and therefore exactly when this script is most likely to be called. Reporting a conflict there would be a false alarm on a perfectly healthy PR. The value is surfaced as data and left for the next check.

**No retry loop was added for it.** GitHub's documented approach is to re-poll until it resolves, but `watch` already re-reads the field on every poll and `fetch` is cheap to re-run — a bounded-retry parameter would be an invented constant solving a problem the existing polling already solves. If a definitive answer is ever needed inside a single `fetch`, that is the point to add it, with a sourced bound.

## No extra round trip

`isDraft`, `mergeable` and `mergeStateStatus` live on the same `pullRequest` object as the head commit, so they are selected in the existing head-commit query rather than a query of their own. `_fetch_latest_commit_date` becomes `_fetch_head_state`, and `build_fetch_result` still makes **seven** `gh` calls, not eight — which matters because those seven are sequential and sit on `watch`'s 90-second poll path.

Verified against this repository's own API (2026-08-30): `mergeStateStatus` needs no preview `Accept` header on `gh api graphql`.

## Design notes

- `mergeStateStatus` is reported as data but drives **no blocker of its own**. Its `DRAFT` and `DIRTY` values restate `isDraft` and `mergeable`; its remaining values (`BLOCKED`, `BEHIND`, `UNSTABLE`, `HAS_HOOKS`) describe merge readiness, not reviewability. Only two conditions actually stop reviews, so only two produce blockers.
- Both it and `mergeable` stay plain strings rather than enums, so a state GitHub adds later reaches the caller as data instead of failing validation on a PR that is otherwise fine. Covered by a test.
- `has_outstanding_work()` is untouched on purpose: a draft PR can still carry unresolved threads that need answering, and a reviewable PR with nothing outstanding is still nothing to act on. Reviewability explains an empty result set; it never creates or suppresses work. Covered by a test.

## SKILL.md

Step 1's "never treat an empty `unresolved` array as nothing-to-do" guidance now enumerates `reviewability.blockers` alongside the three counts, saying explicitly that a non-empty `blockers` means the empty result set is *expected* and the fix is on the PR rather than in the review queue. One gotcha added covering `watch` and the `UNKNOWN` case.

## Verification

```
$ uv run pytest .agents/skills/receiving-pr-reviews/scripts/ -q
50 passed in 3.05s

$ uv run ruff check --no-fix .agents/skills/receiving-pr-reviews/scripts/
All checks passed!

$ uv run ruff format --check .agents/skills/receiving-pr-reviews/scripts/
4 files already formatted

$ uv run ty check .agents/skills/receiving-pr-reviews/scripts/
All checks passed!
```

`prek run --files <the five changed files>` — every applicable hook passed.

42 → 50. New tests: ready PR (empty `blockers`), draft PR, conflicting PR, both-at-once, `UNKNOWN` asserting no false blocker, an unrecognized future `mergeStateStatus`, `watch` carrying the block on a `timed_out: true` result, and `has_outstanding_work()` staying independent of blockers. The draft case is unit-tested only — the repo had no open draft PR to smoke-test against; the conflicting case was verified live against #3338.

## Relationship to the other open PRs

Independent of #3340. Touches the same files as #3341 but different regions — whichever merges second may need a small rebase, no design conflict either way. Written against `main` as it is today, so its models are plain `BaseModel`; if #3341 lands first, the three new ingress fields should inherit its `_GitHubResponseModel` (they are all `bool`/`str`, so strict mode needs no timestamp exemption).


Jamie-BitFlight:

Jamie-BitFlight:
Rebased onto `main` (now carrying #3341 and #3340) and applied the follow-up this PR's own body flagged.

## Conflicts and how each was resolved

Three files conflicted; `pr_review_threads.py` and `SKILL.md` auto-merged. No design conflict — #3341 makes `gh` timeouts caller-supplied, this PR adds fields to an existing query. The two are orthogonal in intent and only collided on shared signature lines, so every resolution is the union of both sides.

| Conflict | Resolution |
|---|---|
| `pr_review_models.py` — #3341 rebased `ForcePushEvent` onto `_GitHubResponseModel` at the same anchor where this PR inserts three new classes | Kept both. New ingress models inherit the strict base (below); `ForcePushEvent` keeps #3341's base. |
| `pr_review_gh.py` — both sides rewrote `_fetch_latest_commit_date`'s signature line: #3341 changed it to `gh_timeout: float | None`, this PR renamed it to `_fetch_head_state` returning `PullRequestHeadState` | Union: `def _fetch_head_state(owner, repo, pr, *, gh_timeout: float \| None) -> PullRequestHeadState`. |
| `pr_review_gh.py` — `build_fetch_result`'s call block: #3341 threaded `gh_timeout_budget(deadline, gh_timeout)` through every call, this PR swapped one call for `_fetch_head_state` and `_latest_revision_at` | Union: #3341's two-argument budget on all seven calls, this PR's head-state call and revision helper. |
| `test_pr_review_threads.py` — the renamed helper's unit test | `pr_review_gh._fetch_head_state("o", "r", 1, gh_timeout=None)` — both sides' intent. |
| `test_pr_review_threads.py` — both sides appended a test block at the same anchor | Kept both blocks: #3341's strict-ingress and timeout-bound tests, then this PR's reviewability tests. |

## The three new fields now come in strict

`PullRequestHeadState` and its nested `HeadCommitsConnection` inherit `_GitHubResponseModel`, so `isDraft` must be a real boolean and `mergeable`/`mergeStateStatus` real strings. All three are `bool`/`str`, so **none needs the `GitHubTimestamp` `strict=False` relaxation** — that exemption stays limited to timestamp fields, where it exists because these models validate `json.loads` output in Pydantic's Python mode and a strict `datetime` would reject GitHub's ISO-8601 string. Not widened.

Proved rather than asserted: two cases added to the existing `test_ingress_models_reject_a_coerced_producer_shape` parametrization — `isDraft: "false"` and `mergeable: 1` are both rejected.

`Reviewability` itself is derived from an already-validated `PullRequestHeadState`, so it is an output shape and stays a plain `BaseModel` alongside `UnresolvedThread` and `FetchResult` — the same rule `_GitHubResponseModel`'s own docstring states.

## Preserved

- Every #3341 fix intact — `gh_timeout_budget(deadline, gh_timeout)` on all seven calls, `run_gh`'s caller-supplied bound, split exception handlers, `min` constraints, unbounded first fetch.
- `mergeable: "UNKNOWN"` still produces **no blocker**.
- Still **seven** `gh` calls, not eight — the three fields still ride the existing `pullRequest` query.

## Verification

```
$ uv run pytest .agents/skills/receiving-pr-reviews/scripts/ -q
66 passed in 3.82s

$ uv run ruff check --no-fix .agents/skills/receiving-pr-reviews/scripts/
All checks passed!

$ uv run ruff format --check .agents/skills/receiving-pr-reviews/scripts/
4 files already formatted

$ uv run ty check .agents/skills/receiving-pr-reviews/scripts/
All checks passed!
```

`prek run --files <the five changed files>` — every applicable hook passed. No failures at any point.

56 (`main`) → 66. Both sides' suites run green together.

Live smoke, post-rebase — and the first one caught the `UNKNOWN` path in production, which is the design decision above working exactly as intended:

```jsonc
// this PR, mid-recompute after a push
{"is_draft": false, "mergeable": "UNKNOWN", "merge_state_status": "UNKNOWN", "blockers": []}

// #3338, genuinely conflicting
{"is_draft": false, "mergeable": "CONFLICTING", "merge_state_status": "DIRTY",
 "blockers": ["conflicting: reviews will not run until the merge conflicts are resolved"]}
```

PR #3388, opened 2026-09-03
feat(git-history-recon): skill-local exclusion file for manifest churn

Jamie-BitFlight:
## Summary

- Add `assets/exclude-patterns` inside the `git-history-recon` skill — a gitignore-style file (glob per line, `#` comments) shipped with the skill itself, not read from the target repo, so its filtering travels with the skill on every install.
- Every default pattern is grounded in a verified cause: the `.claude-plugin`/`.codex-plugin`/`.cursor-plugin` `plugin.json` files and `marketplace.json` trace to `auto_sync_manifests.py`'s `_plugin_manifest_paths()`, and `uv.lock`'s churn was confirmed dependabot-only by sampling actual commits. The npm/yarn/pnpm lockfile entries are explicitly flagged in-file as unconfirmed-by-analogy (this repo has none to sample against).
- Wired in via git's `:(glob,exclude)` pathspec magic, applied only to Pipelines 1 (Hotspots), 2 (Bug Magnets), and 7 (Newly Added Files) — the only pipelines with file-path output. Pipelines 3–6 (commit/author-based) are untouched.
- New EC-7: missing/empty exclusion file degrades to zero filtering rather than failing.
- Report header always states the exclusion count explicitly (including the zero case) — filtering is never silently invisible to whoever reads the report.
- Mermaid flowchart, pipeline prose, and References section all updated in lockstep, per the skill's own "diagram is the authoritative procedure" convention.

## Test plan

- [x] `uvx skilllint@latest check ./.claude/skills/git-history-recon/` — exit 0. The one warning (SK006, skill body over the 4400-token complexity threshold) is pre-existing (5668 tokens before this change) and advisory-only; this change adds to it (6967 tokens) but introduces no new finding.
- [x] Independently re-ran the live pathspec test against this repo's real history: `git log --name-only --since='1 year ago' --format='' -- . ':(glob,exclude)**/.claude-plugin/marketplace.json'` — confirmed `marketplace.json` (840 hits unfiltered) is absent from filtered output while every other file's count is unchanged.
- [x] `prek run` on both changed files — all hooks pass (ruff-format, markdownlint, skilllint validator, conventional commit).


Jamie-BitFlight:
Closing as superseded by #3415 — a multi-angle review found 7 confirmed issues in this PR's exclude-patterns feature (contradictory report-header formats for N=0, a Mermaid condition weaker than EC-7's own detection, a bare/ambiguous EXCLUDE_PATHSPECS placeholder risking shell glob-expansion, an inconsistent SKILL_DIR vs ${CLAUDE_SKILL_DIR} token, an undisclosed pathspec-scoping change vs. Pipelines 3-6, and frozen empirical counts violating this repo's No-Derived-Data rule). #3415 carries the same feature with all of these fixed.

PR #3393, opened 2026-09-03
fix(fastmcp-creator): update connections.py for mcp 2.x API renames

Jamie-BitFlight:
## Summary

`Python / ty` was failing on `main` (found while checking PR #3391's CI before merging). Root cause verified: `origin/main`'s committed `uv.lock` at `ba4d312f5` already pins `mcp==2.1.1` — an upstream `mcp` package release, unrelated to any of this session's other work. mcp 2.x renamed three things `connections.py` used:

- `mcp.client.streamable_http.streamablehttp_client` → `streamable_http_client`
- `Tool.inputSchema` → `Tool.input_schema`
- `streamable_http_client(headers=...)` removed entirely — headers now go through an `httpx2.AsyncClient` passed as `http_client=`

## Test plan

- [x] `uv run -q --locked ty check .` (CI's exact command) passes clean on the full repo
- [x] `uv run ruff check` clean
- [x] Grepped the repo for the same broken patterns elsewhere — none found

### DIFF
--- a/plugins/fastmcp-creator/.claude-plugin/plugin.json
+++ b/plugins/fastmcp-creator/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "fastmcp-creator",
   "description": "Build FastMCP 3.x Python MCP servers \u2014 covers provider/transform architecture (including CodeMode, Tool Search, and server-level transforms), component versioning, session state, authorization (MultiAuth, PropelAuth, connection-pooled token verifiers), evaluation creation, Pydantic validation, async patterns, STDIO and HTTP transports, nginx reverse proxy deployment, background tasks, Prefab Apps UI, security patterns, client SDK usage, testing, deployment, and migration from FastMCP v2. TypeScript is a legacy reference only and is not updated for v3.",
-  "version": "4.3.8",
+  "version": "4.3.9",
   "author": {
     "name": "Jamie Nelson",
     "url": "https://github.com/bitflight-devops"
--- a/plugins/fastmcp-creator/.codex-plugin/plugin.json
+++ b/plugins/fastmcp-creator/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "fastmcp-creator",
-  "version": "4.3.5",
+  "version": "4.3.6",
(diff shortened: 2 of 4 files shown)


Jamie-BitFlight:
Correction to this PR's description: the claim that "origin/main's committed uv.lock at ba4d312f5 already pins mcp==2.1.1" was checked against the wrong environment and is false. `origin/main`'s actual locked versions at that commit are `mcp==1.28.1` and `fastmcp==3.2.4` (verified directly via `git show origin/main:uv.lock`). mcp 1.28.1's `Tool` model still exposes `inputSchema` (camelCase) — confirmed via `Tool.model_fields.keys()` against this branch's own locked environment.

Pushed a follow-up commit that:
- Fixes `import httpx2` → `import httpx` (plain typo, also used at the `AsyncClient(...)` call site)
- Reverts `tool.input_schema` → `tool.inputSchema` (the actual field name on the currently-locked mcp version)
- Keeps the `streamable_http_client`/`http_client=` change as-is — mcp 1.28.1 already ships both the old and new function names, so that part of the original change is valid regardless of version

Verified `uv run -q --locked ty check .` (CI's exact command) and `uv run --locked ruff check` both pass clean against this branch's actual environment, not a different checkout's.

Jamie-BitFlight:
Correction to my own previous comment: I was wrong there too. Root cause, fully verified this time:

GitHub Actions' \`pull_request\` trigger checks out the synthetic merge of the PR head against the *current* tip of \`main\`, not a fixed base. Between my first and second CI checks, an unrelated PR (#3392) merged to \`main\` and bumped \`main\`'s own \`uv.lock\` to \`mcp==2.1.1\` — which has a real transitive dependency on a real PyPI package named \`httpx2\` (not a typo), and whose \`Tool\` model uses \`input_schema\` (snake_case), and whose \`streamable_http_client\` genuinely requires \`httpx2.AsyncClient\`.

The **original, unmodified code in this PR was already correct** for what \`main\` has now — it only failed CI initially because \`main\` hadn't caught up yet. My follow-up commit inverted it to match this branch's own (at-the-time-stale) lockfile, which broke it against current \`main\`.

Fixed by:
1. Reverting to the original \`connections.py\` (my "fix" commit is gone from history).
2. Rebasing this branch onto current \`origin/main\` so its own lockfile also resolves \`mcp==2.1.1\`, matching what CI actually tests against.

Verified \`uv run -q --locked ty check .\` and \`uv run --locked ruff check\` both pass clean post-rebase, against the branch's own now-current lockfile — not a different checkout's.

Apologies for the churn — should be correct now.

Issue #3403, opened 2026-09-04
fix: sam_task append_section concatenates onto stale content instead of replacing on re-run


Jamie-BitFlight:
## Description

`sam_task(action='update', append_section=..., section_content=...)` (backed by `sam_schema/writers/yaml_writer.py:append_section`) is append-only: when a task is reset to `not-started` and re-run (e.g. after a fix loop resets a review task per `dh:multi-perspective-review`/`dh:complete-implementation`'s recursive follow-up handling), the second run's `append_section` write appends its new block after the first run's stale block under the same heading, rather than replacing it.

Reproduction (two independent occurrences in one session): plan P6b323f17 (a `dh:multi-perspective-review` QG run for #3182). T3 (quality review) was reset to `not-started` and re-run after a REJECT verdict was fixed; the re-run's `append_section("Review Results", ...)` call concatenated its new APPROVE verdict after the stale REJECT verdict already in `context-notes`, leaving two JSON objects back-to-back under one `## Review Results` heading (unparseable as a single JSON block). The synthesizer agent hit the identical bug independently on T5 (`Punch List` section) during its own reset-and-retry.

Impact: a section written by `append_section` cannot be safely re-written by re-running the same task — every consumer that parses that section as a single JSON block (e.g. `dh:review-synthesizer` reading `Review Results`, or anything reading `Punch List`) either fails to parse or, worse, could silently pick up stale content mixed with fresh content. In this instance, the synthesizer agent caught it correctly (by re-reading immediately after write rather than trusting `updated: true`) and worked around it via `set_fields_json` to overwrite the whole field — but that workaround requires the caller to already suspect the bug and know the exact field name backing the section.

User-provided context: the fix likely belongs in `append_section` itself — either add a `replace_section`-equivalent mode (mirroring `backlog_groom`'s own `replace_section` parameter, which already solves the identical problem for backlog items) or have callers whose section content is a single structured block (JSON/YAML) use `set_fields_json` directly instead of `append_section`. Exact contract is design work for grooming.

## Context

- **Source**: Discovered twice during dh:multi-perspective-review QG dispatch for #3182 (plan P6b323f17, tasks T3 and T5)
- **Priority**: P2
- **Added**: 2026-09-04
- **Research questions**: None

Jamie-BitFlight:
## Description

[...]

Issue #3402, opened 2026-09-04
fix: sam_task claim_task swallows ContentConflictError into a misleading "wrong status" warning


Jamie-BitFlight:
## Description

`ContentTaskProvider.claim_task` in `plugins/development-harness/sam_schema/core/backends/content.py` (~184-198) swallows a genuine `ContentConflictError` (an optimistic-concurrency conflict raised when multiple agents write the same shared plan content concurrently) into a plain `False` return. `dh_core/operations.py` (~959-971) then reports this as `"Task is 'X', not 'not-started'"` — a status-mismatch message — even when the task is genuinely still `not-started` and the real cause was a concurrent-write race, not a status conflict.

Reproduction (from the reporting agent): claiming task T4 on a plan with multiple concurrently-active sibling tasks (T1-T3 running in parallel on the same plan under `dh:multi-perspective-review`) returned `claimed:false` with the misleading status message on the first two attempts, while `sam_task(read)` and the CLI `plan claim` both independently confirmed the task was actually still `not-started`. A third attempt succeeded once the concurrent writers quieted.

Impact: a caller that takes the warning at face value (rather than retrying) could wrongly conclude the task is unavailable/already claimed and abandon it, when the correct action is a simple retry.

User-provided context: the fix likely belongs in `claim_task`'s exception handling — a caught `ContentConflictError` should be surfaced distinctly from an actual status mismatch (e.g. a retry-signaling return value or a distinguishable exception/error message), rather than collapsed into the same `False`/generic-message path. Exact contract is design work for grooming.

## Context

- **Source**: Discovered incidentally during dh:multi-perspective-review T4 (accessibility) dispatch for #3182's quality gate
- **Priority**: P2
- **Added**: 2026-09-04
- **Research questions**: None

Jamie-BitFlight:
## Description

`ContentTaskProvider.claim_task` in `plugins/development-harness/sam_schema/core/backends/content.py` (~184-198) swallows a genuine `ContentConflictError` (an optimistic-concurrency conflict raised when multiple agents write the same shared plan content concurrently) into a plain `False` return. `dh_core/operations.py` (~959-971) then reports this as `"Task is 'X', not 'not-started'"` — a status-mismatch message — even when the task is genuinely still `not-started` and the real cause was a concurrent-write race, not a status conflict.

[...]

Issue #3400, opened 2026-09-04
fix: interop SKILL.md Step 3 ABORTs on backlog_add error even when the item was still stored


Jamie-BitFlight:
## Description

`plugins/development-harness/skills/interop/SKILL.md` Step 3 (lines 87-95) has a mermaid flow that calls `backlog_add` and branches `CheckError{Response has error key?}` → `Yes: ABORT` (prints the error, does not capture the new issue number, does not write `**Backlog item:** #N` into the plan file).

After #3182's fix, `backlog_add`/`add_item` can return a populated `error` on a call where the item WAS still stored locally — the "GitHub issue creation attempted and failed" branch (`item_ref`/local storage succeeds, `result["error"]` carries the reason). Read literally, this flow's ABORT branch would discard that successfully-stored item and never write its reference into the plan file — the same contradiction `create/start.md` had before #3182's T5 task fixed it there.

Not in #3182's original Impact Radius (interop/SKILL.md was not found as a consumer of `backlog_add`'s response contract during grooming); discovered by dh:service-docs-maintainer while implementing T5's equivalent fix in a sibling doc.

User-provided context: the fix pattern is already established in `create/start.md` Step 5 (post-#3182) — branch on a populated `error` separately from "no item created," report the error, but still capture the issue number/reference and proceed rather than aborting. The same treatment likely applies here, but the exact wording is design work for grooming, not prescribed here.

## Context

- **Source**: Quality vigilance concern from #3182
- **Priority**: P2
- **Added**: 2026-09-04
- **Research questions**: None

Jamie-BitFlight:
**Closed** (superseded). **Reference**: #3417 
Moot: #3417 (merged) superseded PR #3404's approach specifically because it set backlog_add's `error` (singular) field on the stored-but-issue-creation-failed case, violating FallibleToolResponse's established contract. #3417's shipped design instead routes that failure through the pre-existing `Output.errors` (list) channel via `out.record_error()`, and never sets `error` for the non-fatal case. interop/SKILL.md's Step 3 check (which keys on `error`, singular) is therefore never triggered by this case at all — the contradiction this item described no longer exists under the shipped design. Confirmed directly in #3417's PR body: "interop/SKILL.md needs no update, since it keys on error (singular), which this fix never sets for the non-fatal case."

Jamie-BitFlight:
## Description

[...]

Issue #3395, opened 2026-09-04
fix: backlog_add and CLI add don't catch ContentUnavailableError from the beads backend


Jamie-BitFlight:
## Description

Neither the MCP tool `backlog_add` (server.py:1404) nor the CLI `backlog add` command (sam_schema/backlog.py, `except BacklogError`) catches `ContentUnavailableError`. The beads backend's `create_beads_issue_for_item` raises `ContentUnavailableError(ContentProviderError(Exception))` on issue-creation failure (see `TestAddItemBeadsBackend`, test_backlog_core_operations.py:565-650) — a sibling exception hierarchy to `BacklogError`, not a subclass of it (confirmed by direct source read of models.py: `BacklogError(Exception)`, `ContentProviderError(Exception)`, `ContentUnavailableError(ContentProviderError)`).

On a beads-backed project, a failed `create_beads_issue_for_item` call therefore propagates as an unhandled exception through both `backlog_add` (MCP) and CLI `add`, instead of returning the clean `{"error": ...}` response shape both call sites already produce for `BacklogError`.

Impact: an agent or CLI user on a beads-backed project calling `backlog_add`/`backlog add` when beads issue creation fails gets a raw unhandled-exception traceback instead of a structured error response — inconsistent with the GitHub-backend path's error handling.

Discovered as an out-of-scope finding while architecting the fix for #3182 (backlog_add's GitHub-path silent-failure defect) — see plan/architect-backlog-add-silent-failure.md concerns section. Not reproduced live (would require a beads-backed project with a failing beads issue-creation call to trigger); the exception-hierarchy mismatch itself is confirmed by direct source read, not a hypothesis.

## Context

- **Source**: Agent task - auto-derived from input
- **Priority**: P2
- **Added**: 2026-09-04
- **Research questions**: None

Jamie-BitFlight:
## Description

[...]

PR #3404, opened 2026-09-04
fix(backlog-core): surface GitHub issue-creation failure reason from backlog_add

### LINKED ISSUE #3182: fix: backlog_add returns an empty item_ref when issue creation fails, and the reason reaches no caller
## Description

`backlog_add` can complete with `item_ref: ""` — no GitHub issue created — while returning a success-shaped result. The caller receives a populated dict and no error.

Observed in CI on `main`, run 32624257966, job "Python / E2E live tests":

```
plugins/development-harness/tests/test_live_validation.py:196: in test_l1_add_with_real_issue
    assert item_ref.startswith("#"), f"Expected item_ref like '#N', got: {item_ref!r}"
E   AssertionError: Expected item_ref like '#N', got: ''

plugins/development-harness/tests/test_live_validation.py:318: in test_l11_resolve_alternative_end
    l11_issue_num = int(create_result["item_ref"].lstrip("#"))
E   ValueError: invalid literal for int() with base 10: ''
```

Two symptoms, one defect. The remaining lifecycle tests then skip on the `l1_ok` sentinel, and two more pass only because they assert types rather than values.

Why the cause is still unknown, which is the substance of this item: `operations._try_create_github_issue` (around line 1377) returns `None` on either `try_get_github(...) is None` ("GitHub unavailable") or a caught `GithubException`/`BacklogError` ("Issue creation failed: {e}"). Both are recorded only via `out.warn(...)`. The reason therefore lives in `result["warnings"]`, and neither test assertion includes it, so roughly four weeks of CI logs contain no information about why creation failed.

This is the failure shape `.claude/rules/silent-failure-prevention.md` prohibits directly: a write operation that does not report what changed, and a branch on an input condition whose fallback neither acts nor errors. An agent calling `backlog_add` cannot distinguish "issue created" from "issue not created, item exists locally only".

Not caused by any recent change. Verified independently: every Code Quality run on `main` back through 2026-08-20 failed, including commits well before the recent backlog_core work (58df37b85, 79962289b, d22c524e6, 0a2ca9177). A subagent sampled e2e job conclusions on `main` on 2026-07-28, 08-06, 08-10 through 08-13, and every run from 08-14 to 08-23 — all failure, zero successes, so at least four weeks.

The symptom text changed once without the defect changing. Before 2026-08-18 it read `Expected item_ref in result, got: [...]` with `'reference': 'p1-mcp-test-<uuid>-live-test-item'` — the local-only fallback. PR #3160 made `item_ref` always present, so it became `got: ''`. Same bug throughout.

Ruled out as causes, with evidence: not a flake (20+ consecutive identical failures; the fixture uses a fresh tmp dir, a per-run UUID prefix and `force=True`); not rate limiting or auth rejection (step completes in 12.4s, no 403 or 429); not leftover state; not run contention. The job's token carries `Issues: write`.

User-provided context: a subagent's suggested cheapest first move is to include `result["warnings"]` and `result["errors"]` in the test assertion message, so the next CI run reports the actual cause — but that is a diagnostic aid, and the defect to fix is that a failed create is not surfaced to the caller at all. Whether the right contract is raising, an explicit status field, or something else is design work for grooming.

## Context

- **Source**: Diagnosed from the E2E live tests failure on main, 2026-08-23
- **Priority**: P1
- **Added**: 2026-08-23

Jamie-BitFlight:
## Summary
- `_try_create_github_issue`/`_try_create_backend_issue_ref` now return `(value, error)` tuples instead of collapsing a failure reason to `None`; `add_item` conditionally sets `result["error"]` only on the genuine creation-failure branch, never on the tolerated GitHub-unavailable branch (#2999 regression guard preserved).
- CLI `backlog add` exits non-zero when the result carries a populated `error`.
- `test_live_validation.py`'s two `item_ref` assertions now surface the error reason instead of an opaque empty-string/type-conversion failure.
- `create/start.md` documents the new three-outcome contract (created / local-only-and-fine / attempted-and-failed), including a fix (commit `51823112c`) distinguishing "item stored, issue creation failed" from the pre-existing "item never stored" (duplicate/validation) error path — caught by an independent quality-review pass before this PR was opened.

Fixes #3182

## Quality gates run
- TN verification (7/7 acceptance criteria PASS, no regressions)
- Multi-perspective review: Security APPROVE, Performance APPROVE, Quality APPROVE (after one fix cycle), Accessibility SKIP (no UI files)
- Independent code review: PASS, 0 blocking findings
- Feature verification (goal-backward): VERIFIED, no gaps
- Integration check: all producer/consumer/test/doc wiring CONNECTED
- Documentation drift audit: 0 findings

## Follow-ups filed (out of scope for this PR, tracked separately)
- #3395 — beads backend `ContentUnavailableError` not caught by `backlog_add`/CLI `add`
- #3400 — `interop/SKILL.md` has the same stored-but-errored ABORT contradiction this PR fixed elsewhere
- #3402 — `sam_task` `claim_task` swallows `ContentConflictError` into a misleading status warning under concurrent claims
- #3403 — `sam_task` `append_section` concatenates onto stale content instead of replacing on task re-run

## Test plan
- [x] `test_backlog_core_operations.py` (160 passed)
- [x] `test_frontend_parity_ops.py` default + `-m integration` (48 + 2 passed)
- [x] `test_backlog_cli.py` (10 passed)
- [x] `backlog_core`/`sam_schema` full suites (879 passed)
- [x] ruff + ty clean on all touched files

https://claude.ai/code/session_01GHZ9qtoiRw8MW6C5gAuCXh

### DIFF
--- a/plugins/development-harness/.claude-plugin/plugin.json
+++ b/plugins/development-harness/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dh",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
-  "version": "11.0.15",
+  "version": "11.0.17",
   "author": {
     "name": "Jamie Nelson",
     "url": "https://github.com/bitflight-devops"
--- a/plugins/development-harness/.claude-plugin/plugin.json
+++ b/plugins/development-harness/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dh",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
-  "version": "11.0.17",
+  "version": "11.0.19",
   "author": {
     "name": "Jamie Nelson",
     "url": "https://github.com/bitflight-devops"
--- a/plugins/development-harness/.codex-plugin/plugin.json
+++ b/plugins/development-harness/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "dh",
-  "version": "10.0.15",
+  "version": "10.0.17",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
   "author": {
     "name": "Jamie Nelson",
--- a/plugins/development-harness/.codex-plugin/plugin.json
+++ b/plugins/development-harness/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "dh",
-  "version": "10.0.17",
+  "version": "10.0.19",
   "description": "Language-agnostic development process harness implementing the Stateless Agent Methodology (SAM) 7-stage pipeline with ARL human touchpoint model and Voltron-style language plugin composition. Provides orchestration, workflows, planning, verification, and testing methodology that any language plugin can compose with.",
   "author": {
     "name": "Jamie Nelson",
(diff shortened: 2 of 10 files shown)


Jamie-BitFlight:
Closing as superseded by #3417 — code review found the tuple-threading approach and new singular `result["error"]` field violate `FallibleToolResponse.error`'s established contract and bypass `Output`'s existing `errors: list[str]` channel (already used this exact way elsewhere in this file, e.g. `get_sam_tasks`). #3417 fixes the same underlying defect by routing through that existing channel instead — same behavior fix, much smaller diff, no schema/contract change.

PR #3527, opened 2026-09-13
feat(research-curator): detect unresolved repo-path citations in analysis files

Jamie-BitFlight:
## Summary

- A corpus survey found ~83 `research/insights/` and `research/utilization/` files whose "Already covered" verdicts rest on citations to repository paths that do not exist: `.claude/rules/` (moved to root-level `rules/` on 2026-09-04, and in several cases never existed as `.claude/rules/` at all) and a family of skill names that has never existed here (`swarm-operations`, `swarm-patterns`, `swarm-orchestrating`, `swarm-spawning`, plus additional never-existing names found during measurement: `swarm-from-markdown`, `orchestrating-swarms`).
- Adds a `repo_path_unresolved` warning-severity check to `validate_research.py`, scoped to `research/insights/` and `research/utilization/` gap-analysis files (a scoping change — `_NON_ENTRY_DIRS` previously excluded these directories from validation entirely; they're now collected and validated by a lighter path that runs only this check, not the entry-template structural checks).
- The check only fires on lines carrying an existing-state assertion ("Already covered", "**Local system**:", "**Caller**:", etc.) and excludes lines with a negation or aspirational marker ("does not exist", "to be created", "new skill", etc.) — this scoping and gating was arrived at empirically: an unscoped/ungated version of the same check produced an **~86% false-positive rate** on entries (external tools' own paths, not this repo's), so entries are deliberately left to `entry-review-rubric.md`'s existing manual Gate 4 "Path exists" review instead. The final design measured **zero false positives** on the corpus sample reviewed while building it, and found **142 unresolved citations across 67 files** on a full real-corpus run.
- Full design rationale, scope boundary, and measurement writeup is in `references/validation-rules.md`'s new "Repository Path Citations" section.
- Does **not** fix any of the ~83 files — the check comes first, remediation is a separate follow-up once the detector is trustworthy (per explicit instruction).

## Test plan

- [x] `uv run ruff check` / `uv run ruff format --check` / `uv run ty check` on the modified script — all clean
- [x] `uv run pytest tests/research_backlinks/` — 124 passed (100 pre-existing + 24 new in `test_repo_path_citations.py`), no regressions
- [x] `npx markdownlint-cli2` on the updated `validation-rules.md` — clean
- [x] Ran the validator against the real `research/` corpus (506 files) and manually audited a sample of flagged hits to confirm the measured false-positive rate


Jamie-BitFlight:
Closing unmerged.

This check and #3528's Gate 4 are two implementations of one job — verifying claims a research entry makes about this repository — and neither is wired to a consequence. Shipping both pays twice for a check that acts zero times.

Measured against the two clusters it was built for: 40% and 46% recall. Its only wired consumer is the pre-commit hook invoked without `--json`/`--verbose`, which prints a bare warning count with no filename, path, or line. `validate_analysis_file` computes `has_errors` over issues that are unconditionally `warning`, so status is always `pass` and the check cannot block anything.

The defect it targeted is real and recurring — PR #3391 renamed `.claude/rules/` to `rules/` and PR #3422 removed the swarm skills, between them breaking 99 citations across 83 files. Those citations are repaired in #3531. Retested against recurring drift rather than one-time cleanup, this check's shape is still wrong for it: the marker gating that produces the 40%/46% recall is what a drift checker would need to delete, not tune.

Salvage audit found nothing on this branch worth keeping — every supporting function exists only to serve `repo_path_unresolved`, and the two changes that look like fixes to pre-existing code repair functions the branch itself introduced one commit earlier. See #3532.

PR #3528, opened 2026-09-13
refactor(research-curator): gate entries on what the owner needs from them

Jamie-BitFlight:
## Why

The entry review rubric ran six gates over a research entry. None of them measured whether the entry was worth having.

Measured this session:

- A stratified sample of **29 entries** against the current rubric: **0 APPROVE, 208 defects**.
- Mechanical gates corpus-wide (n=328): **81% fail Gate 1**, on **1409 warnings and 0 errors**.
- Gate 5 (did the analysis engage this repo) fails **79%** and did not move under any contract change tried.
- 211 repo claims checked, **74 defective (35%)** — the one gate whose failures cost a reader a wasted session.
- Rule 2a's 91% failure rate is an artifact: the template *required* a "Key Statistics" section until `a54eef252` (2026-07-08). Entries written after that fail it 0%.
- Gate 6's keyword scan over-predicts ~3.5x (35% flagged, 10% adjudicated).

The repo owner's framing, verbatim:

> the content all says the url it came from, if any local agent is doing tasks where they need to check if theres a known technology or system to achieve something, then i expect them to read the canonical source and not the research summary anyway.

> creating single page references that assist with knowing the state of the art, and applying lessons from it to this repo

> remove everything that doesnt change the outcome

If nobody acts on the entry's account of the subject, every gate auditing that account's fidelity changes no outcome. What is left is what the entry alone provides.

## What an entry is for

1. **Deliver a reader to the canonical source, dated.** The entry is the pointer and the reason to follow it — never the authority.
2. **Say what this repo should do about the subject, in terms true here.** Only this repo's files can settle that, and a wrong claim costs a session.

## The rubric

**Gate 1 — pointer integrity.** Validator errors zero; source URL present and resolving (`curl -L`, status recorded); a verification date; no bare "doesn't support X" talking the reader out of going to the source. Decides the verdict.

**Gate 2 — repo claims.** Every claim about this repository, per claim, against the file it names: names something here → exists → described correctly → gap is real → signal runs. The old Gate 5 is folded in as step 1, so "would be equally true of any repository" is now a falsifiable per-claim test rather than a section-level judgment that never moved.

**Verdict: `USABLE` | `UNUSABLE`**, decided by Gate 1 alone. An entry with a live dated pointer is worth having with repairs outstanding, so defects no longer withhold its README row — only an entry that points nowhere does. Previously `APPROVE` required `DEFECTS: 0`, which is why 0 of 29 approved and why worth-having entries were being hidden from the index over a wrong path in a proposal file.

## Mechanical vs judgment

| Concern | Owner |
|---|---|
| Required sections, empty sections | `validate_research.py` error severity, via the Validation Gate — an entry with errors never reaches review |
| Header/freshness/access-date/cross-ref-absent warnings | `validate_research.py`, **reported not scored** |
| Repo-path existence in `insights/`/`utilization/` | `validate_research.py` `repo_path_unresolved` (PR #3527) — each hit is a confirmed Gate 2 step-2 defect. Referenced conditionally, so this PR is correct before and after #3527 lands and does not touch the script |
| Markdown formatting | `prek`, Post-Actions step 4 |
| Cross-reference symmetry | `check-backlinks --fix`, Post-Actions step 2 |
| Source URL resolution, claim adjudication, generic-analysis test | Judgment, in the rubric |

No change to `validate_research.py`, so this does not conflict with #3527.

## Removed, and why each changes no outcome the owner wants

| Removed | Why |
|---|---|
| `fix_research_formatting --check` as a gate | `prek` already fixes formatting in Post-Actions. A formatting nit never made an entry not worth having |
| `check-backlinks` as a gate | Post-Actions step 2 repairs symmetry deterministically *before* review. This also deletes the "repair reciprocity first" ordering constraint the Entry Review section had to carry solely to stop the gate failing every entry on a defect the same run was about to fix |
| Fidelity Rule 1 per-claim traceability | Unauditable after the fact, and the reader goes to the source |
| Fidelity Rule 2 (exact counts) | Nobody acts on the entry's figures. Kept as a writing standard, at zero review cost |
| Fidelity Rule 2a as a review check | Rescoped to *gathering*. Existing statistics in existing entries stay — a figure in a finished entry is not a defect, not a `--fix` target, not a reason to edit |
| Fidelity Rule 4 (confidence map) as a review check | A `high` vs `medium` on a section does not change whether the reader opens the source. Kept as a writing requirement, so SKILL-GOALS goal 1 still holds at production |
| Gate 3 depth, all four sections | "Present-but-thin is a defect" is the direct contradiction of "read the canonical source, not the research summary." Kept as writing standards |
| Gate 6 hallucination triggers | The scan over-predicts ~3.5x; its findings inside the gated sections are already caught by Gate 2, and AGENTS.md already routes agent-output review to `/hallucination-detector:hallucination-audit` |
| `APPROVE` requiring `DEFECTS: 0` | Made the verdict uncorrelated with worth-having, and hid good entries from the index |

Fidelity Rule 3 survives as a Gate 1 check, in one direction only: a bare nonexistence claim is the one fidelity failure that stops the reader reaching the source.

## Added

The depth requirement that was missing for the one section the review gates. "Relevance to Claude Code Development" now has an explicit bar — name a local path and open it, state what it does today, tie the change to a mechanism the entry documented from the source, name the signal — with the instruction to prefer three items that name files over ten that do not, and to write "No application to this repository found in the sources reviewed" rather than pad. This is the section behind the 79% and 35% numbers, and it previously had no standard at all.

`research-insight-extractor` and `research-utilization-assessor` were checked and left alone: both already require path-grounded, file-read proposals. Their output was failing because nothing held them to it, which is what Gate 2 now does.

## SKILL-GOALS

All six goals are FROZEN and none regress. Goals 2, 4, 5 and 6 are preserved or strengthened; goal 3 gets cheaper per-entry review.

**One tension to report, not resolved unilaterally:** goal 1 requires "confidence levels attached to each claim." Under the owner's decision, a per-section confidence level changes no reader outcome, because the reader goes to the canonical source regardless. This PR keeps confidence as a *production* requirement in the standards and template (goal 1 intact) but stops *auditing* it. If the goal intends the audit too, it and the owner decision cannot both hold — flagging rather than changing a frozen goal.

## Verification

- Dry-ran the new Gate 1 against `research/agent-infrastructure/kernel-sh.md`: validator `errors 0, warnings 3`, `metadata.source_url` present, `https://www.kernel.sh` → HTTP 200, `metadata.verified: 2026-02-22` present → PASS. The three warnings are exactly the reported-not-scored kind, and the run exposed that the validator's freshness key aliases miss `verified`, so both Gate 1 presence rows now instruct reading the entry rather than trusting the warning.
- `uv run prek run --files` on all four changed files — clean.


Jamie-BitFlight:
Closing unmerged. The diagnosis was right and the fix was aimed one layer too high.

Its finding stands: the six gates all measured transcription fidelity, so the rubric could not fail an entry that was scrupulously accurate and useless, and 0 of 29 sampled entries passed. But rewriting the gates cannot fix it, because the entries name almost nothing real about this repo — 292 of 331 Relevance sections contain no repository path at all.

The cause sits upstream. That section had no writing instructions anywhere in the skill, and `extraction-methodology.md` had no phase that reads this repository, so a gate demanding repo-grounded claims was unsatisfiable by construction. #3530 fixes that with a Repo Anchor Pass before writing, where a zero-match search counts as valid evidence.

Also closing because this PR's Gate 2 and #3527's regex check both open every repository path a proposal names — one job implemented twice, neither wired to a consequence.

Salvage in #3532: the record that the hallucination triggers are local by decision, and the fix stopping creation targets being scored as missing files. The rest was either already on `main` in stronger form or repaired hazards this branch introduced.

PR #3530, opened 2026-09-13
feat(research-curator): anchor Relevance items to real repo paths

Jamie-BitFlight:
Scope: **the entry production side.** The analysis-agent redesign was started here and has been handed back — see "Handed to the plan" below.

---

## The defect

`## Relevance to Claude Code Development` had **no writing instructions**. Every mention across the skill and agents is review-side or validation-side:

| Location | Role |
|---|---|
| `references/entry-template.md:113` | three placeholder bullets |
| `references/entry-review-rubric.md:69,:84` | two gates that judge it |
| `scripts/validate_research.py:54` | required-section list |
| `references/extraction-methodology.md:33` | section-name enum only |
| `.claude/agents/research-curator.md:217` | `--review` mode only |

Two compounding causes:

1. **Wrong subject.** The heading says "Claude Code Development" and the placeholder says "integrate this with Claude Code" — the product, not this repository. An agent writing "Claude Code skills that perform web research need reliable browser automation" has obeyed the template precisely. The 292 unanchored entries are following the instruction, not failing it; the gates then fail them for not naming files in a repo the template never asked about.
2. **No phase read this repo.** `extraction-methodology.md` extracts exact passages from the external source before abstracting — correct for every section whose material lives in the source. This section's material lives *here*, and no phase looked. A template field asking "which file does this bear on" has no answer unless something earlier went and looked.

Measured: of 331 entries with the section, 39 contain a repo-relative path, 292 contain none.

## Phase 1c — Repo Anchor Pass

Unconditional, between code analysis and writing. Extract from this repo the way Phase 1 extracts from the resource: `ls` the real names; derive 3-6 search terms from the entry's own extracts (mechanisms, not the brand name, which by definition will not appear here); `grep -ril` each over `plugins/ .claude/ rules/ docs/ AGENTS.md`; read up to six matched files and quote one line each. Six Reads is the budget; unsearched terms are reported, not dropped.

**Zero matches is an anchor, not a dead end.** This is what makes the standard satisfiable for every research subject rather than only the lucky ones — the search command is the evidence, and any reader can re-run it.

Scope is explicit in-file: this pass finds paths and quotes lines. It does not duplicate the downstream analysis agents' gap assessment, confidence scoring, or backlog work.

## Anchored item shape

```
- **{capability}** -> `{repo-relative path}`
  - Today: "{exact line read from that path}"
  - Change: {the specific edit this suggests}

- **{capability}** -> nothing in `{scope searched}`
  - Today: `grep -ril "{term}" {scope}` -> 0 matches
  - Change: {what would have to exist here first}
```

`Change:` has three passing outcomes: a specific edit, `none — {path} already covers it`, or `none — out of scope ({why})`. **A standard with no honest exit is what produced force-fitted prose in the first place** — this is a different shape, not a stricter version of the old one. The subject is disambiguated in a note rather than by renaming the heading (see below).

`--rerun` re-runs the anchor pass unconditionally, since anchors go stale independently of the resource.

## Applied by hand to three entries of different shapes

All three change materially; anchors found in 2-4 tool calls each.

- **`msgspec.md`** (library, five generic "Primary Use Cases") — anchor pass finds `AGENTS.md:252` *"Structured data → Pydantic, not dataclass/TypedDict"*, `python3-typing/SKILL.md:37` *"### 3. Python 3.11+ with Pydantic"*, and `grep -ril msgspec` → **0 matches**. Becomes one `out of scope` finding (the repo has ruled on this shape; msgspec would contradict a standing convention, not extend it), one concrete skill edit (the lane-selection table has no zero-copy branch), one verified absence.
- **`system-prompts-ai-tools.md`** (meta/corpus) — "could strengthen Claude Code skill SKILL.md files" becomes `skill-creator/SKILL.md`, whose `allowed-tools` documentation is a permission gate with no ordering/fallback guidance. And `git grep -il refusal -- plugins/ .claude/skills/ .claude/agents/ rules/ docs/ AGENTS.md ':!research/'` returns exactly one tracked file, `plugins/plugin-creator/skills/mission-statement/SKILL.md` — a structural gap the vague version could never surface.
- **`vibe-kanban.md`** (sharpest) — two of its four "Patterns Worth Adopting" propose things the repo already has. Worktree isolation → `rules/commit-cadence-and-worktrees.md` already mandates it, *including* the committed-state constraint vibe-kanban's model does not address. SSE → already the documented MCP transport in three files. Both become verified `already covers it`.

## Analysis agents: two discovery-agnostic edits only

The redesign of `research-insight-extractor` and `research-utilization-assessor` was started in this branch and **reverted** (commit 3). What survives holds under any discovery design:

- **`**Absence evidence**` field** required behind any "no local system does X" claim, plus: every path a proposal names must have been opened first.
- Same two rules added to the utilization assessor's boundaries, with the note that most of this repo's skills live under `plugins/*/skills/`, not `.claude/skills/`.

Grounded in a measured defect: `research/insights/2026-03-10-cocoindex-code-improvements.md:37` asserts *"No skill in `.claude/skills/` or `plugins/` provides semantic code search capability"* and targets creating one, while `plugins/python3-development/skills/semantic-code-search/SKILL.md` and `plugins/python-engineering/agents/semantic-code-search.md` both exist. One `grep -ril "semantic" plugins/` would have prevented it.

## Handed to the plan

Reverted from this PR and owed to the dedicated insight-assessment plan: where overlap discovery happens, what the agents read to do it, the candidate-then-deep-dive shape, per-entry cost across 331 entries, and what the redesign deletes.

One finding worth carrying into it, since it contradicts the working hypothesis: **the twelve-row `Mapping Patterns to Local Systems` table is not stale — all eleven of its paths currently resolve.** It is *incomplete*: twelve domains, and most research subjects fall outside all of them. Auto-generating it would have prevented nothing. Concluding absence from a missing row is the actual mechanism. Separately, the 21-of-32 nonexistent paths are in *entry* Relevance sections, written by the curator, which has no table and had never read the repo — a different population with a different cause, addressed by Phase 1c here.

## Reported upward, not changed

- **The heading is the residual defect.** `## Relevance to Claude Code Development` names the product where the content must describe the repository. Renaming it needs a coordinated change across `scripts/validate_research.py:54`, `entry-review-rubric.md:69,:84`, and 331 entries — out of scope here and held by other agents. The template note is a mitigation, not the fix.
- **Rule 2a violation in an existing insights file**: `2026-03-10-cocoindex-code-improvements.md` cites "337 stars". `research/` is out of scope for this PR.

## Constraints

Six frozen `SKILL-GOALS.md` goals: none edited, none regressed; goal 5 is strengthened. No file under `research/` modified — the three applications above are a dry run. `entry-review-rubric.md`, `entry-quality-standards.md`, `SKILL.md`, `validate_research.py`, `.pre-commit-config.yaml` untouched. Existing statistics untouched; entry length not constrained.

---

**Correction (review follow-up).** This example was originally written with `grep -ril refusal`,
which returns tens of files rather than one: plain `grep` descends into gitignored
`.claude/worktrees/`, so most hits are worktree copies no clone has, and the count drifts per
machine and per hour. The count of one holds only under `git grep` over the scope
`entry-template.md` defines. Phase 1c now specifies that command; see the
`fix(research-curator): anchor Phase 1c to git grep and make it discriminating` commit and its
follow-up `fix(research-curator): correct false gitignore claims and the term/Read conflation`.


Jamie-BitFlight:
## Second-round review fixes (2026-09-14)

Commits `c1527319d` (F8-F12 + the two `4c06ddab8` defects + the 298-warning items), `1cdbe3ec3` (merge main).

| Finding | Disposition |
|---|---|
| F8 — search command manufactures absences from any subdirectory | **Fixed at the root, not with a precondition.** Reproduced first: from `<worktree>/plugins` the documented command returns nothing with exit 1 and no message; from the root it returns `plugins/plugin-creator/skills/mission-statement/SKILL.md` with exit 0. A stated precondition an agent can skip silently is a weak fix, so the command is now root-anchored by construction: `:/` on every pathspec plus `--full-name` so the printed path stays repo-relative and usable as the anchor record's path. Applied in `extraction-methodology.md`, the A2 record template, and `.claude/agents/research-curator.md`. |
| F9 — gate green-lights hallucinated paths | **Fixed.** New `relevance_anchor_path_missing` resolves every cited path against the checkout root (`repo_root_for`, walking to `.git` — a file in a linked worktree, hence `.exists()` not `is_dir()`). **11 corpus entries cite 14 non-existent paths**, including all four the review named. Deliberately carries **no** date exemption — predating Phase 1c excuses an entry from having anchors, not from citing a file that is not here. Globs/placeholders are skipped as shapes. Where no `.git` is found the run emits `relevance_anchor_paths_unchecked` rather than passing, per `rules/silent-failure-prevention.md`. |
| F10 — Gate 4 Rule 5 non-executable | **Fixed by making the field reach the artifact**, not by weakening the rule. `entry-template.md`'s present-anchor item shape now carries a `Term:` line copied from the anchor record's `Term matched:`, and the rubric checks against it. A missing `Term:` is itself recorded as a defect, so the gate can never land on `NOT RUN` for a field the entry was required to write. |
| F11 — budget contradiction | **Fixed.** Six Reads is now the only budget and per-capability Reads are explicitly uncapped, which makes the rejection path at `:200-201` reachable and consistent. `.claude/agents/research-curator.md` carries the same wording instead of only the first half. |
| F12 — `':!research/'` described as load-bearing | **Confirmed and removed.** Measured: 344 results with it, 344 without. `research/` lies under none of the six scope paths. Dropped, with the reason it needs no exclusion stated in its place. |
| `4c06ddab8` — hardcoded counts in `research-utilization-assessor.md` | **Reverted** to the pre-correction qualitative phrasing ("a path recalled rather than listed is usually wrong"), per AGENTS.md's rule against restating directory-derived values. |
| `4c06ddab8` — false grep rationale | **Confirmed false and replaced with the measurement.** On `refusal` in the primary checkout: **73** files under plain `grep -ril` over `plugins/ .claude/ rules/ docs/ AGENTS.md`, **3** under `git grep` over that same scope, **1** under the narrowed command — and over the six paths alone **both commands return the same single file**. The 73→3 is the worktree descent; the 3→1 is the narrowing. Both effects are now stated where they actually apply, instead of the worktree claim being attached to the six-path scope where it is false. |

### The 298-of-331 warning

Severity unchanged, as reviewed. Both named items addressed:

- **Enforcement.** The prose rule "must-fix only for entries a run just wrote" had nothing enforcing it, while `check_cross_references` ten lines below already carried the mechanical date exemption that would. Applied that precedent as `RELEVANCE_ANCHOR_EXEMPT_BEFORE`, keyed on the same Research Date / Last Verified resolution. The shared date logic is extracted to `exempt_by_date` and both checks now call it. **`relevance_unanchored` drops 298 → 63.**
- **Scope.** Rerun Mode's inclusion is left as-is and its cost is now stated plainly in `validation-rules.md`: a refresh updates Last Verified and therefore gates the entry, which is consistent rather than accidental, because Rerun Mode's own graph runs ReAnchor on every pass — an entry that comes out of a refresh still unanchored is a skipped ReAnchor, not a legacy artifact. The consequence (first refresh does Phase 1c work it never did; if skipped, "refreshed with issues" and no analysis fan-out) is written down.

### Confirmed-correct items

Left undisturbed as directed: the `--rerun` graph, Gate 5, Gate 4 Rules 6 and 7, Gate 2 Rule 3's extension, `d3e4297cf`, and the corrected tracked-file claim.

### ty diagnostic

The reported `too-many-positional-arguments` at `:890`/`:907` does not reproduce: `uv run ty check` from inside `.claude/worktrees/rc-relevance` reports "All checks passed!". It was reading an intermediate state — `check_relevance_anchor_paths` briefly took four parameters while drafting, and the fourth (`reference_date`) was dropped before commit when the path check was given no date exemption. Definition and both call sites take three at the tip.

### Merge rather than rebase

#3529 and #3532 landed on `main` while this was in flight. Merged `origin/main` in (`1cdbe3ec3`) rather than rebasing, because this branch is already pushed and rewriting it would need a force-push, which is not approved here. One conflict in `validate_file`: #3529 removed `_check_formatting_suggestions` while this branch added calls around it — resolved by taking main's removal and keeping this branch's calls. No stale reference remains in the script or `validation-rules.md`.

### Checks after

`tests/research_backlinks/` **122 passed** (5 new tests pinning the path check, 2 pinning both sides of the date cutoff); `ruff check`, `ruff format --check`, `uv run ty check` all clean; corpus run **331 passed / 0 errors**, `relevance_unanchored` 63, `relevance_anchor_path_missing` 15.

Issue #3546, opened 2026-09-14
feat: detect the sandbox GraphQL block and fall back to the REST API


Jamie-BitFlight:
## Description

GitHub GraphQL is blocked in Claude Code agent sessions. Every `dh` code path that reaches
GraphQL fails there, and the failure is currently indistinguishable from a generic GraphQL error.

Add detection for this specific condition, then fall back to REST under two rules:

1. **A GraphQL call whose data one REST call can also get or set** — swap the function to REST
   outright. GraphQL buys nothing there.
2. **A GraphQL call that replaces multiple REST calls** — keep GraphQL first, and add an
   efficient asynchronous REST fallback that runs only when detection fires.

GraphQL exists in this codebase because it cuts request count. That reason stands, so rule 2
preserves it rather than discarding GraphQL everywhere.

## Reproducibility

```bash
gh api graphql -f query='{viewer{login}}'
curl -s -o /dev/null -w '%{http_code}' -X POST https://api.github.com/graphql -d '{"query":"{viewer{login}}"}'
```

Both return HTTP 403. The response body carries a fixed message:

```text
GitHub GraphQL is not available from Claude Code sessions; use the REST API
(gh api repos/{owner}/{repo}/...). ...
documentation_url: https://docs.anthropic.com/en/docs/claude-code/github-actions
```

## Root-Cause Analysis

`backlog_core/gh_client.py:490` holds `_graphql_request`, the single chokepoint every GraphQL
path goes through. On line 513 it catches `GithubException` and re-raises
`BacklogError(f"GraphQL request failed: {exc}")`.

That wrapper discards the HTTP status and flattens the distinguishing message into free text, so
no caller can tell "GraphQL is blocked in this environment, retry over REST" apart from
"this query is malformed" or "the network is down". Detection has to happen at the
`GithubException` boundary, where `exc.status` is still 403 and the message is still intact.

The file already carries `_is_not_found_error(error: BacklogError)` at line 527, which inspects a
wrapped message for a 404-equivalent condition. A sibling detector fits that established shape,
but should read the status rather than the flattened string.

## Impact Radius

**Systems Inventory**

| System | Path | Change |
|---|---|---|
| Chokepoint | `backlog_core/gh_client.py:490` `_graphql_request` | Detect the blocked-GraphQL 403 and raise a distinct error type |
| Protocol surface | `backlog_core/backend_types.py:266-330` `GitHubExtras` | 13 `*_graphql` members to classify and route |
[...]

PR #3552, opened 2026-09-14
feat(backlog-core): classify a GraphQL refusal, and stop swallowing it in listings

Jamie-BitFlight:
**Stacked on #3551** — base branch is `claude/shared-github-client-519`, not `main`. Review that one first.

Relates to #3546. Layers 1 and 2 of 5.

> **Scope note:** this PR was opened as Layer 1 only. Layer 2 (commit `bccfd02`) was pushed to the
> same branch rather than a new stacked branch. Splitting it back out needs a force-push to this
> branch, which is not something I do without being asked. Say the word and I will split it.

---

### What was wrong

Some sandboxes serve GitHub's REST API and reject every GraphQL request with HTTP 403:

```console
$ gh api graphql -f query='{viewer{login}}'
GitHub GraphQL is not available from Claude Code sessions; use the REST API ... (HTTP 403)
```

`_graphql_request` at `backlog_core/gh_client.py` caught `GithubException` and re-raised
`BacklogError(f"GraphQL request failed: {exc}")`. That discards the status and flattens the
distinguishing message into free text, so no caller can tell these three apart:

- this environment refuses GraphQL — a REST path would work
- this query is malformed — retrying will not help
- this token lacks a scope — REST will not help either

### What changed

`GraphQLUnavailableError` names the first condition. It subclasses `BackendUnavailableError`, which
subclasses `BacklogError`, so every existing `except BacklogError` site still catches it and no
handler changed behaviour in Layer 1.

The detector requires **403 AND a message marker**. A 403 alone is deliberately not the signal — a
token missing a scope returns 403 too, and widening the detector to all 403s would classify a real
permission error as "retry over REST", which cannot work. Tests cover that boundary directly: a
scope error and a rate-limit error both stay generic. The marker text is recorded verbatim beside
the constant, with the date it was observed.

---

### What was wrong

`batch_fetch_statuses` returns a mapping, and an empty mapping is already a real answer: *no item
carries a status*. The refusal was caught at `gh_client.py:1465` and converted into that same empty
mapping, so every listing rendered blank statuses and named no cause.

The probe that pinned it — note the request fires even for an empty item list:

```console
cached items: 0
supports_batch_status_fetch: True
--- calling batch_fetch_statuses with the cached (empty) item list ---
Request POST /graphql failed with 403: Forbidden
    returned: 0 entries
```

### What changed

- `GraphQLUnavailableError` now propagates out of `batch_fetch_statuses`. Every other failure keeps
  its existing empty-map fallback, so an offline read still serves the cache rather than raising —
  the offline continuity `backlog_core/ARCHITECTURE.md` describes is deliberate and is preserved.
- `list_items` catches the refusal, records a warning naming the cause, and continues. The listing
  still renders; it is no longer silent about why the statuses are blank.
- An item list carrying no numeric issue reference returns early without a request. There is no
  lookup key to resolve, so the round trip built a map no caller could read.

---

## Verification

- 22 unit tests for the classification (message extraction across the body shapes PyGithub produces,
  the 403-plus-marker boundary, status parametrisation, exception chaining, subclass relationships).
- 11 unit tests for the surfacing (refusal propagates; generic `BacklogError`, `GithubException` and
  an unreachable backend all keep the empty-map fallback; the pointless-query skip; the warning is
  present on a degraded listing and absent on a healthy one).
- Live check against the blocked API, with a real item:

  ```console
  RAISED GraphQLUnavailableError: GraphQL is unavailable in this environment:
  GitHub GraphQL is not available from Claude Code sessions; use the REST API
  ```

  It previously returned `{}` here.
- `uv run pytest plugins/development-harness/backlog_core/tests plugins/development-harness/tests_backlog`
  — 1300 passed.
- `ruff check`, `ruff format`, `ty check` clean.

## Still open on #3546

- `backlog view --selector "#519"` reports `ItemNotFoundError` for an issue that exists — its own
  swallow is not yet traced.
- A listing served from a cache that has never synced is still indistinguishable from a listing of a
  genuinely empty backlog (`count: 0`, no warning). The CLI path has no `SyncState` at all, so this
  needs a different fix site than the MCP server's `_build_sync_state_block`. Next layer.
- Layer 3 adds `databaseId` to the comment queries; Layer 4 swaps the six single-call functions to
  REST; Layer 5 adds async REST fallbacks for the six multi-call ones.

> Note: the `claude-review` check is expected to fail on this repo right now — its Codex credits are
> exhausted, per the bot's own comment.


Jamie-BitFlight:
@codex review

Issue #3541, opened 2026-09-13
fix: Eleven research entries cite fourteen repository paths that do not exist


Jamie-BitFlight:
## Description

The `relevance_anchor_path_missing` check added in PR #3530 resolves paths cited in an entry's "Relevance to Claude Code Development" section against the checkout root. On the corpus as merged it reports 11 entries citing 14 paths that do not resolve.

These are pre-existing, not introduced by that PR — the check is what makes them visible. They were left unrepaired deliberately: entry review Gate 4 Rule 1 says not to repair a near-miss on the writer's behalf, because a corrected path attached to unverified prose converts a visible broken link into an invisible false claim. That failure mode was confirmed three times during today's corpus repair, where a path was made to resolve while the content it pointed at had been replaced.

Two of the 14 are near-misses of real files: `.claude/rules/proactive-fix-gate.md` where the real path is `rules/proactive-fix-gate.md`, and `.claude/mcp.json` where the real path is `.mcp.json`. Both are the `.claude/rules/` drift pattern that PR #3391's rename created and PR #3531 repaired across 83 other entries.

The remaining 12 have not been classified. Each needs the same treatment the corpus repair applied: check git history to establish whether the path was correct when written, then either repoint it with the content re-verified, record it as superseded with the commit that removed it, or mark the claim unsupported. Do not repoint without re-reading what the target says now.

**User-provided context**: the check is warning severity and gated by `RELEVANCE_ANCHOR_EXEMPT_BEFORE`, so these do not fail any commit or gate today. Related: #3534 covers the fact that entry review cannot express drift at all — it has one outcome for a path that does not open, which attributes repository refactors to the research agent that wrote the entry.

## Context

- **Source**: Session observation — surfaced by the new relevance_anchor_path_missing check added in PR #3530 (merged as ad543e995)
- **Priority**: P2
- **Added**: 2026-09-13
- **Research questions**: None

Jamie-BitFlight:
## Description

The `relevance_anchor_path_missing` check added in PR #3530 resolves paths cited in an entry's "Relevance to Claude Code Development" section against the checkout root. On the corpus as merged it reports 11 entries citing 14 paths that do not resolve.

[...]

Issue #3540, opened 2026-09-13
fix: Repo content recommends TeamCreate, which no longer exists in Claude Code as of v2.1.178


Jamie-BitFlight:
## Description

63 files in the tree reference `TeamCreate` (measured with `git grep -l TeamCreate`), and several recommend using it. They split four ways: 43 under `research/`, 11 generated artifacts (`.json`/`.html`/`.yaml`), and 9 hand-written prose files, of which only a few are genuine recommendations. The call no longer exists: `plugins/agent-orchestration/skills/delegate/references/harness-notes/claude-code.md` records that as of Claude Code v2.1.178, naming a teammate on an `Agent` call while `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS` is enabled spawns it directly, with no separate setup step, and that the `swarm-*` skills documenting the old flow were retired in favour of `parallel-work`.

Confirmed positive recommendations include `plugins/plugin-creator/skills/plugin-creator/SKILL.md` at lines 48 and 815. This is plugin content, not research corpus, so it was outside the scope of the PRs that surfaced it.

An agent following any of these writes a delegation against a primitive that is gone. The failure is silent at authoring time — the instruction reads as valid and only fails when executed.

This is the same drift class the research corpus repair addressed today: PR #3391 renamed `.claude/rules/` to `rules/` and PR #3422 removed the swarm skills, between them breaking 99 citations across 83 files that were all correct when written. Those were repaired in PR #3531. The `TeamCreate` references are the same shape in a different corpus, and nothing detects them.

Two distinct questions the grooming should separate: what each of the hand-written references should say now, and whether anything should detect this class of drift in plugin content going forward. Related open items on the detection half: #3534 (entry review cannot express path drift) and #3535 (claims are checked after the irreversible filing step, not before).

**User-provided context**: the count was recorded during review of PR #3531 as 20 files (remeasured at 63 on 2026-09-15; see the corrected figure above), correcting an earlier figure of 19 that appeared in a banner across 36 research entries. That banner text has since been corrected in the merged PR. Duplicate detection matched five unrelated items on generic keywords (#3037, #3054, #3223, #3262, #3266); none concern this subject and the item was filed with force.

## Context

- **Source**: Session observation — round-two review of PR #3531 (merged as 2a733bebd)
- **Priority**: P1
- **Added**: 2026-09-13
- **Research questions**: None

Jamie-BitFlight:
## Description

[...]

Issue #3534, opened 2026-09-13
fix: Entry review Gate 4 cannot express path drift, so it attributes repo refactors to research agents


Jamie-BitFlight:
## Description

`.claude/skills/research-curator/references/entry-review-rubric.md` Gate 4 step 1 has exactly one outcome for a repository path that does not open: a defect recorded against the entry or proposal that named it. The rubric gives the reviewer no way to record that the path was correct when written and was broken later by a repository change.

A cited path fails to open for two different reasons, and Gate 4 can express only one of them. A path may never have existed — the corpus holds one such case, `swarm-orchestrating`, a name inversion of the real `orchestrating-swarms`. A path may also have been correct when written and then moved or removed by a later repository change:

- `.claude/rules/` existed until PR #3391 (`bf4dcd876`, 2026-09-04) renamed it to root-level `rules/` — 59 citations.
- The swarm skills existed under `.claude/skills/` until PR #3422 (`4e1e73bd6`) removed them.
- `backlog-tools-administrator` existed until PR #3330 (`20c84f3b5`) removed it.

Consequence: Gate 4 systematically misattributes ordinary repository refactors to the research agents that wrote the citations, and records a defect against work that was accurate at the time. Because a Gate 4 defect drives `REQUEST CHANGES`, which withholds the entry's README row (SKILL.md Post-Actions step 1), a rename elsewhere in the repo can keep correct entries out of the index. The defect class recurs on every rename or removal — it is not a closed population that goes empty after the current repair.

What the gate must be able to distinguish, and does not:

1. A path that never existed — a defect against the writer.
2. A path that existed when written and was moved or removed since — repository drift, a repair to schedule against the citing file, not a finding against the agent that wrote it. Recording it should name the change that moved it where that is determinable.
3. A path named as a place to create something — not a defect for being absent at all. (Addressed separately in PR #3532's second commit; listed here because all three outcomes live in the same step.)

Interacts with the filing-gate consequence item: both concern when a claim is checked relative to when it is acted on. That item checks the live tree at write time, which is the only point at which drift has not happened yet; this item concerns the audit that runs afterwards, by which time it has.

[...]

PR #3558, opened 2026-09-14
fix(skills): describe the teammate spawn mechanism that exists

Jamie-BitFlight:
Refs #3540.

`TeamCreate` was removed from Claude Code in v2.1.178. Naming a teammate on an `Agent` call while `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS` is enabled now spawns it directly, with no separate setup step — recorded in `plugins/agent-orchestration/skills/delegate/references/harness-notes/claude-code.md`.

## Measured populations

`git grep -l TeamCreate` returns **63 files**, not the ~20 the item claimed. They split four ways:

| Population | Count | Treatment |
|---|---|---|
| `research/` | 43 | Untouched — another agent owns that corpus |
| Generated artifacts (`.json`, `.html`, `.yaml`) | 11 | Untouched — workflow-extractor outputs, not hand-editable |
| Hand-written prose | 9 | Read each; 3 repaired, 6 left as description |
| — of which genuine recommendations | 3 files | Repaired below |

## Repaired

Three skills carried instructions an agent would act on:

- **`plugins/plugin-creator/skills/plugin-creator/SKILL.md`** (lines 48, 815) — the two confirmed positive recommendations. Both now name the mechanism that exists: pass a teammate name on the `Agent` call. Line 48 points on to `agent-orchestration:parallel-work` "Persistent teams", which already routes to the harness notes for the mechanics, so the version fact lives in one place rather than being copied here.
- **`plugins/plugin-creator/skills/skill-sync/SKILL.md`** (line 62) — a dispatch step steered by prohibition ("not sequential, not `TeamCreate`"). Now states the requirement positively: dispatch the three `Agent()` calls in one turn so they run concurrently.
- **`plugins/development-harness/skills/kage-bunshin/SKILL.md`** (line 411) — a parenthetical ("no `TeamCreate` call needed") implying the call is an alternative a reader might otherwise reach for. The sentence carries its full meaning without it.

## Left as description

A description of what a system once did, or of what a parser scans for, is not a defect. None of these would lead a reader to attempt the call:

- `plugins/agent-orchestration/skills/delegate/references/harness-notes/claude-code.md` — the authoritative record of the removal, and the source this PR cites.
- `docs/dh-backend-beads/comparison-research/04-dh-sam-system.md` — a dated snapshot of the `/implement-feature` dispatch loop.
- `plugins/development-harness/docs/plans/prd-workflow-extractor.md` (5 hits), `docs/graph-schema.md`, `docs/workflow-layers/COVERAGE.md`, `docs/workflow-layers/KNOWN_ENTITIES.md` — the workflow extractor's pattern catalog and coverage notes, describing strings the extractor scans source for.

## Regeneration: not needed by this change

The 11 generated artifacts were not hand-edited. Their `TeamCreate` strings are extraction inputs and captured findings from the last extraction run, and none quotes the kage-bunshin line this PR changed (verified by grep). Regeneration runs through the `dh:meta-workflow-graph-refresh` skill, whose Tier 1 / Tier 2 process lives in `plugins/development-harness/docs/workflow-layers/COVERAGE.md`; note that the `dh-extract-file.js` step that process names is not present in this tree. A refresh would produce a large diff for no correctness gain here, so it is not run in this PR.

## Two corrections made to #3540 itself

The item is about references that drift; its own body had both a wrong count and a drifted citation, and both are now fixed with `gh issue edit`:

1. **Count** — "roughly 20 files" corrected to the measured 63, with the four-way split recorded.
2. **Citation** — the item cited `plugins/plugin-creator/SKILL.md`, which is not a file. The real path is `plugins/plugin-creator/skills/plugin-creator/SKILL.md`.

## Deliberately left open

#3540 asks a second question: whether anything should detect this class of drift in plugin content going forward. **This PR does not answer it and builds no detector.** That question belongs to grooming, joined with the same question in #3534 and #3541, so #3540 stays open after this merges.

## Verification

- `git grep -l TeamCreate -- ':!research/' ':!*.json' ':!*.html' ':!*.yaml'` — no remaining hand-written file recommends the call.
- `uv run prek run --all-files` — passes.
- `uv run ty check .` — passes.

The `plugin.json` version bumps in the diff are the marketplace-version-sync pre-commit hook's own output.


Jamie-BitFlight:
Pre-rebase disposition plan for `fix/3540-teamcreate-recommendations` onto current `origin/main` (merge base `cd8537607`):

- `plugins/development-harness/.claude-plugin/plugin.json`: DROP stale branch bump `11.1.10`; retain target `11.1.36`.
- `plugins/development-harness/.codex-plugin/plugin.json`: DROP stale branch bump `11.1.10`; retain target `11.1.36`.
- `plugins/development-harness/.cursor-plugin/plugin.json`: DROP stale branch bump `11.1.10`; retain target `11.1.36`.
- `plugins/development-harness/skills/kage-bunshin/SKILL.md`: MERGE target grooming-command migration and branch removal of the obsolete `TeamCreate` suggestion; regions are disjoint.
- `plugins/plugin-creator/.claude-plugin/plugin.json`: DROP branch bump because target already contains the exact `14.12.6` change.
- `plugins/plugin-creator/.codex-plugin/plugin.json`: DROP branch bump because target already contains the exact `14.12.6` change.
- `plugins/plugin-creator/skills/plugin-creator/SKILL.md`: REWRITE the shared recommendation lines to preserve the branch’s explicit named-teammate mechanics while retaining target’s runtime-path portability paragraph.
- Branch-only, NO_CONFLICT: `plugins/plugin-creator/skills/skill-sync/SKILL.md`.

This plan was derived from file-level diffs on both sides before running `git rebase`.

Issue #3536, opened 2026-09-13
fix: Rule 2a is enforced retroactively, so refreshing a pre-2026-07-08 entry trips a defect for obeying its own template


Jamie-BitFlight:
## Description

`.claude/skills/research-curator/references/entry-review-rubric.md` Gate 2 scores Rule 2a as: "Any such figure anywhere in the entry, including inside a badge, a quoted README passage, or a 'Key Statistics' section that should not exist."

That section was required by the entry template until commit `a54eef252` (2026-07-08, PR #2723) deleted the `Key Statistics` block — `GitHub Stars`, `Downloads/month`, `Contributors` — from `.claude/skills/research-curator/references/entry-template.md`. Verified in that commit's diff. Every entry authored before that date therefore carries the section by construction, because the template it was written against demanded it.

Consequence: a `--rerun` refresh of any pre-2026-07-08 entry trips a Rule 2a defect for having obeyed the template in force at the time. A Gate 2 defect drives `REQUEST CHANGES`, which withholds the entry's README row (SKILL.md Post-Actions step 1), so the entry drops out of the index for a rule that did not exist when it was written. The rule reads as a statement about entry content when what it governs is an agent's gathering behaviour.

The repo owner has ruled that existing statistics stay. Any rule about popularity statistics governs an agent gathering them — the prohibition binds at gather time, applies whether or not the repository is in the session's authorized GitHub scope, and admits no fallback source (web search, package registry, a README badge). It does not bind the existing files. A figure in a finished entry is not a defect, not a `--fix` target, and not a reason to edit or withhold that entry.

Two places state the rule and both need to agree: the fidelity rule in `entry-quality-standards.md` (the writing contract) and the Gate 2 row in `entry-review-rubric.md` (the audit). Today only the audit half asserts retroactivity, so fixing one file alone leaves the contradiction in place.

Same false-defect class as #3534 and the creation-target fix already shipped in PR #3532: a gate scoring text that was correct when written.

**User-provided context**: the original wording is in closed PR #3528, commit `0f3a15bf6`, in the `entry-quality-standards.md` Rule 2a rewrite. It was left out of the salvage PR because it changes whether a finished entry is withheld from the index, which is verdict mechanics rather than salvage. The historical fact and its citation (`a54eef252`, 2026-07-08) are the durable part.

## Context

- **Source**: Session observation — salvage review of closing PRs #3527 and #3528 (PR #3532)
- **Priority**: P2
[...]

PR #3559, opened 2026-09-14
fix(research-curator): stop entry gates 2 and 4 faulting the writer for repo drift

### LINKED ISSUE #3534: fix: Entry review Gate 4 cannot express path drift, so it attributes repo refactors to research agents

Jamie-BitFlight:
Fixes #3534 and #3536 together — they edit the same two files and are the same defect class: the entry review rubric produces a defect against the *writer* for a condition the writer did not cause, and any defect withholds the entry's README index row (SKILL.md Post-Actions step 1). A correct entry drops out of the index because the repository changed around it.

## #3534 — Gate 4 step 1 could not express path drift

Step 1 had one outcome for a repository path that will not open: a defect against the entry that named it. It now resolves into three, settled by `git log --all --full-history -- {path}` plus the claim's own wording:

- **Never existed** — the log has never seen the path. Defect against the writer.
- **Existed and moved** — the log returns the commits that held it. The entry was accurate when written; a repository change since then moved or removed the path. Recorded as a **repair** against the citing file, naming the commit.
- **A place to create something** — unchanged, and now a peer of the other two rather than an aside.

## #3536 — Rule 2a was enforced retroactively

The `Key Statistics` block (`GitHub Stars`, `Downloads/month`, `Contributors`) was required by the entry template until commit `a54eef252` (2026-07-08, PR #2723) deleted it, so every earlier entry carries it by construction. Rule 2a now binds an agent at gather time and leaves figures a finished entry already carries. Stated once, in the `entry-quality-standards.md` contract; the rubric's Gate 2 row defers to it, so the two halves cannot drift apart.

## Repair as a record type

`Repair` joins `Defect` at the top of the rubric: same evidence discipline, plus the change that invalidated the text, counted separately and reported in its own `REPAIRS` block. Repairs leave the verdict where it stood — an entry with repairs and `DEFECTS: 0` is `APPROVE` and keeps its README row. `SKILL.md`'s relay instruction now carries repairs through to the run report.

## Out of scope, deliberately

How this repository should bind a reference to what it referred to, so drift becomes detectable, is a separate open architectural question. No detector is designed or added here; these gates work whichever way that question is answered.

## Verification

- `uv run .claude/skills/research-curator/scripts/validate_research.py main ./research` — 331 of 331 passing.
- `uv run prek run --all-files` — all hooks pass. `uv run ty check .` — all checks passed.
- **Pre-2026-07-08 entry**: `research/ai-research-tools/the-unwind-ai.md` (added 2026-02-19), which carries a `Key Statistics` section with `GitHub stars (awesome-llm-apps) | 95,911` and a forks row. Old Gate 2 row — "a 'Key Statistics' section that should not exist" — makes this a Rule 2a defect on any `--rerun`, failing Gate 2 and withholding the entry's README row (it holds one today, `research/README.md:1260`). New row asks whether *this run* gathered the figure; a rerun gathers nothing, the existing figures stay, Gate 2 passes.
- **Entry citing a moved path**: `research/insights/2026-04-25-wrkflw-improvements.md:35` cites `` `.claude/rules/ci-workflows.md` `` in a claim about what that file stated. The path does not open — `.claude/rules/` was renamed to root-level `rules/` in PR #3391 (`bf4dcd876`). Old step 1 has only the "what is there now" branch available, so this is a Gate 4 defect against the writer. New step 1 runs the log, which returns `bf4dcd876` and earlier commits, and routes it to the **existed and moved** branch: a repair against the citing file naming that commit, Gate 4 PASS, verdict and README row untouched.


Jamie-BitFlight:
Reviewed the full diff and verified the two test cases independently.

`research/ai-research-tools/the-unwind-ai.md:33` does carry a `## Key Statistics` section. `research/insights/2026-04-25-wrkflw-improvements.md:35` does cite `.claude/rules/ci-workflows.md`, that path does not open today, and `git log --all --full-history -- .claude/rules/ci-workflows.md` returns `bf4dcd876` (PR #3391). So the "existed and moved" branch is exercised by a real case, not a constructed one.

Two observations, neither blocking.

**The Rule 2a row asks a question a reviewer cannot always answer.** "Did *this run* leave every star, download, fork, and contributor count out of the entry?" A reviewer auditing a `--rerun` sees the finished entry, not the diff, so it cannot separate what this run gathered from what the entry already carried. The row resolves the ambiguity in the safe direction — figures the entry already carried always stay — so the failure mode is a missed defect on a fresh entry rather than a false one against a correct entry, which is the right way round for this change. The binding gate is `entry-quality-standards.md` at gather time. Worth knowing that the audit-side row is advisory for reruns.

**`REPAIRS:` has no consumer yet beyond the run report.** That is correct for this PR — who acts on a repair is the open architectural question this work was scoped away from, and inventing a queue here would have foreclosed it. Recording it: the grooming discussion that takes up the drift-class question (joined across #3534, #3540, #3541) now has a structured data source that did not exist before this change.

Jamie-BitFlight:
One design gap, small, and worth a follow-up rather than a change here.

Step 1 now resolves a non-opening path through three outcomes, settled by `git log --all --full-history -- {path}` plus the claim's wording. The structure encodes a precedence — the wording picks whether the claim is about what is there now or about a place to create, and the log then splits the first kind into never-existed and existed-and-moved. That precedence is implicit rather than stated.

There is no outcome for a reviewer that cannot decide. A claim whose wording sits genuinely between "the hook in `Z` writes the field" and "integration point: `Z`" has no branch, and the likely failure is that the reviewer picks whichever branch it has seen most often. The correct behaviour there is a distinct undecidable outcome, not a guess weighted by how common each case is.

That matters more than the frequencies suggest. Base rate is not cost per case: the never-existed branch is the rarest in this corpus and it is the only one where a Gate 4 defect against the author is correct. A reviewer that defaults to the common branch when unsure loses exactly the case the gate exists to catch.

Separately — I handed the Lane A prompt a corpus ratio ("of roughly 99 broken citations, exactly one was ever fabricated"). It is not in this PR, and it should not be: the classifier decides per path, so the mix changes no branch, no detection mechanism, and no verdict mapping. Dropping it was right. I have removed it from #3534's body for the same reason.

PR #3570, opened 2026-09-14
fix(backlog-core): stop gh_client swallowing non-refusal GitHub failures

Jamie-BitFlight:
## Summary

Task B1 of the #3546 synthesized plan (`scratchpad/design/synthesis-plan.md`, "Question B — Decided approach"), fixing the largest finding in `B-critique.md` §3.3.

`gh_client.py`'s `try_get_github` folded **any** `GithubException` from `get_repo` into the same `None` a missing token produces — not just a missing-token case. `batch_fetch_statuses` and `view_enrich_from_github` then caught `(BacklogError, GithubException)` broadly and fell back to `{}`/`False`. Only the one GraphQL-refusal marker string was preserved as a distinguishable cause; a network failure, a 500, or a rate limit was indistinguishable from "nothing was wrong; there's just nothing there" — directly contradicting these functions' own docstrings.

- `try_get_github` now returns `None` only for a missing token (a configuration state) and raises `GitHubUnavailableError` for any other `GithubException` from `get_repo`.
- `batch_fetch_statuses` and `view_enrich_from_github` raise `GitHubUnavailableError` instead of swallowing a non-refusal `BacklogError`/`GithubException` into an empty map/`False`.
- `GitHubUnavailableError` is a `BackendUnavailableError` subclass, so `operations.py`'s existing `except BackendUnavailableError` clauses around these two calls (in `list_items` and `view_item`) already catch it correctly — **no new except clause was needed there**.
- The contract change ripples to every other `try_get_github` caller that relied on it collapsing every failure into `None` for a documented local-only fallback: `operations.py`'s `_create_issue_and_update_item`, `_rename_item_title`, `_apply_plan_to_item`, `_try_create_github_issue`; `gh_client.py`'s `probe_backend_status` (whose docstring promises it never raises); and the `_GitHubContentCache` read/write paths in `github_content_migration.py`. Each now catches `BackendUnavailableError` around the `try_get_github` call and preserves its prior graceful-degradation behavior.

## Test changes

- `backlog_core/tests/test_batch_status_refusal.py` and `backlog_core/tests/test_refusal_not_item_missing.py`: updated the pinned assertions that expected a generic `BacklogError`/`GithubException` to fall back to an empty map/`False` — they now assert `GitHubUnavailableError` propagates. The GraphQL-refusal propagation tests and the missing-token fallback tests are untouched. Added a new test in each file exercising the `try_get_github`-level failure directly (the earlier of the two swallow points).
[...]

PR #3572, opened 2026-09-14
fix(backlog-core): stop a degraded status batch from fabricating filter matches

Jamie-BitFlight:
## Summary

Task B2 of the synthesized #3546 implementation plan (`scratchpad/design/synthesis-plan.md`, "B2" section), stacked on B1's PR #3570 (`claude/gh-client-stop-swallowing-3546`).

Fixes the status-filter fabrication bug proven in `scratchpad/design/B-critique.md` §3.1: `_item_derived_status` returned the literal `"needs-grooming"` for *every* numeric-issue item whenever the live status map was empty, with no way to tell "the fetch answered and this item genuinely has no status label" apart from "the fetch never answered at all". Concretely, under a degraded/refused status batch:

- `list_items(status="needs-grooming")` **fabricated** matches — returning items whose true live status was something else entirely.
- `list_items(status="in-progress")` **silently dropped** items that do match, with no signal that the result was unreliable.

B1 already made a degraded/refused status batch fetch raise a typed `BackendUnavailableError` (a `GraphQLUnavailableError`/`GitHubUnavailableError` subclass) instead of silently returning `{}`. `list_items` already catches that exception (verified by reading the current code, not assumed) — but fell through to the same empty `status_map`, reproducing the exact bug.

## Fix

Threaded a `status_map_unavailable` flag from `list_items`'s `except BackendUnavailableError` handler through `_filter_open_items` into `_item_derived_status`, which now returns `None` (never a guessed default) for a numeric-issue item when the flag is set — distinguishing "the map answered, no entry for this item" (legitimate default to `"needs-grooming"`, unchanged) from "the map never answered for anyone" (`None`, matches nothing). A `status=` filter can never equal `None`, so a degraded result excludes those items instead of asserting a match that isn't true.

The existing warning is now filter-aware: when a `status=` filter is active under a degraded map, it names the filter that could not be verified, rather than describing blank status rendering that no longer applies once matching items are excluded outright (so the false-negative case is disclosed, not silent).

## Testing

New file `plugins/development-harness/backlog_core/tests/test_status_filter_fabrication.py`, mirroring `scratchpad/design/proofs/test_critique_proofs.py`'s P1/P2 reproduction shape (adapted to B1's typed-exception API), proves:

- `list_items(status="needs-grooming")` under a degraded map returns 0 items, not the 3 (2 factually wrong) the bug produced.
- `list_items(status="in-progress")` under a degraded map returns 0 items but the result now carries a warning naming the unverifiable filter — no longer silent.
- Unit coverage of `_item_derived_status`'s new discriminator, including that string-ID/issueless items (unaffected by the live map) are untouched.

Validation (all green):

```
uv run ruff format --check .        # 2839 files already formatted
uv run ruff check .                 # All checks passed!
uv run ty check                     # All checks passed!
uv run pytest backlog_core/tests tests_backlog tests -q
  # 4395 passed, 51 skipped, 5 xfailed
```

Relates to #3546

---
_Generated by [Claude Code](https://claude.ai/code/session_01GXD1kzsunY2r3HJjwP6Txx)_


Jamie-BitFlight:
@codex review

---

Jamie-BitFlight:
The "Preserve degradation warnings in count-only responses" finding is already fixed downstream in PR #3575 (task B4 of this same backlog item's plan), which merges `out.warnings`/`out.errors` into the `count_only` response — explicitly covering the "Live status unavailable" warning this and PR #3572 ship. Not duplicating the fix here to avoid two divergent implementations colliding when the stack merges forward; it will land in this PR's history once #3572 → #3574 → #3575 merge in sequence.

The other finding on this PR (missing-token status fetches still fabricating matches) is being fixed directly on this branch.

---

PR #3573, opened 2026-09-14
fix(backlog-core): read through a never-synced GitHub cache once

Jamie-BitFlight:
## Summary

Task A3 of the synthesized implementation plan for #3546 (`scratchpad/design/synthesis-plan.md`, A-critique.md §5 "ALT-5"). Stacks on A2 (`claude/snapshot-unreadable-discoverable-3546`).

`list_items` (`backlog_core/operations.py`) previously only *described* a never-synced GitHub cache's ambiguity (the `778ffc1` warning: a cold cache and a genuinely empty backlog render identically). It never tried to fix it. Now, when the snapshot checkpoint is genuinely `None` — an honest "never synced" signal since A1's checkpoint-honesty fix — and the caller did not already request `refresh=True`, `list_items` triggers exactly one `refresh_local_cache_from_github` attempt before returning results, per the critique's ALT-5: *"it fires once, on a condition that by construction cannot recur after it succeeds, not on every list."*

- Adds a narrow `SnapshotCheckpointProvider` protocol (`backend_types.py`) exposing `has_synced_snapshot()`, implemented by `GitHubBackend`/`_GitHubReconciliation` — deliberately **not** folded into `SyncProvider` itself, since that would force every existing `isinstance(x, SyncProvider)` gate in `operations.py` (the never-synced-cache warning included) to also implement it, silently changing behaviour for any backend or test double that doesn't.
- A failed refresh attempt (no token, offline, still refused) is swallowed — an implicit read-through a caller never asked for must not turn an unaware listing into a hard error — and the existing "cache holds no items" warning still fires afterwards, unchanged.
- A warm cache (`has_synced_snapshot() == True`) never triggers the auto-refresh path.

### Resolving the "fires once" ambiguity

The checkpoint stays honestly `None` on a failed refresh (per A1's semantics), so a *later*, separate `list_items()` call against a cache that never manages to sync will attempt the refresh again. This is deliberate, not a bug: "fires once" is read as *once per `list_items()` call* (never a retry loop within one call, and never on an already-warm cache), not *once for the process lifetime*. Three reasons:

1. The critique's own contrast is "fires once... not on every list" — i.e. distinguishing this from the correctly-rejected per-list-probe alternative (§6.2) that would fire regardless of cache warmth. It never claims a process-lifetime single-shot.
2. A process-lifetime "already tried" flag would be new, unmodeled mutable state outside the file-cache checkpoint this whole track is built to keep honest — and would silently stop offering the fix to a caller whose credentials become valid mid-session.
3. The critique frames a failed attempt as giving *sharper* meaning on each subsequent call ("we tried and could not" vs. "we never tried"), which only holds if each call keeps trying while the cache remains genuinely uninitialized.

## Test plan

- [x] New regression tests in `backlog_core/tests/test_refusal_not_item_missing.py` (`TestColdCacheReadsThroughOnce`), proving: (a) a cold cache triggers exactly one refresh attempt; (b) `refresh=True` doesn't double-fire; (c) a failed refresh doesn't loop and the existing `_EMPTY_CACHE_MARKER` warning still fires unchanged; (d) two separate cold-but-failing calls each attempt once (2 total, no loop within either call); (e) a successful refresh returns real item data, not just a provenance note; (f) a warm cache never triggers a spurious refresh.
- [x] `uv run ruff format` / `uv run ruff check` — clean
- [x] `uv run ty check` — clean
- [x] `uv run pytest plugins/development-harness/backlog_core/tests` — 706 passed
- [x] `uv run pytest plugins/development-harness/tests_backlog` — 626 passed
- [x] `uv run pytest plugins/development-harness/tests` — 3061 passed, 51 skipped, 5 xfailed

Relates to #3546

---
_Generated by [Claude Code](https://claude.ai/code/session_01GXD1kzsunY2r3HJjwP6Txx)_


Jamie-BitFlight:
@codex review

---

PR #3609, opened 2026-09-16
fix(dh): gate the CLI's PYTHONPATH re-exec guard to direct execution only

Jamie-BitFlight:
## Problem

`sam_schema/cli.py` clears `PYTHONPATH` and re-execs itself, so a project's own copy of a dependency cannot shadow the CLI's resolved environment. That guard was top-level module code, so it also ran on `import sam_schema.cli`.

About 15 test modules import that module in-process. With `PYTHONPATH` set in the host process, the import called `os.execve` with the host's own `sys.argv` and replaced the host process.

Measured on main before this fix:

```
PYTHONPATH=/tmp uv run pytest tests/test_cli_active_task.py -q -p no:randomly --no-cov
# every xdist worker crashes, "no tests ran", exit 0
```

## Fix

The guard now runs only when the CLI runs as a script (`__name__ == "__main__"`), and `.pop("PYTHONPATH", None)` keeps its default.

## Checks

- New test `test_importing_cli_module_does_not_hijack_a_foreign_pythonpath_host` runs pytest as a subprocess on a file that imports the CLI module, with `PYTHONPATH` set. It failed before the fix with empty output and exit 0.
- `PYTHONPATH=/tmp uv run pytest tests/test_cli_active_task.py` → 21 passed.
- `uv run pytest tests/test_frontend_parity.py -m integration` → 5 passed, including the two foreign-dependency cases.
- ruff check and ruff format clean.

Refs #3449


Jamie-BitFlight:
CI triage at base SHA `5965c7b8e6c7cb1fa8fd4fe22aa5bbc1f28d42a7`: all four failing categories reproduce unchanged in a detached clean-base worktree. `uv run -q --locked prek run markdownlint-cli2 --all-files` fails on `docs/backend-providers.md`; `uv run -q --locked prek run ruff --all-files` fails on `test_open_pr_search_failure.py`; `uv run -q --locked ty check .` fails on `github_backend.py`; and the CI-listed tests reproduce the same portability/backlog failures. None of those files is changed by this PR. Branch-focused verification passes: `uv run pytest plugins/development-harness/tests/test_frontend_parity.py -q` reports `4 passed`, and `uv run prek run --files` over all five PR files passes every applicable hook. No unrelated baseline changes were added to this focused fix.

Jamie-BitFlight:
## `/code-review high` on `037a5f890`

**The core fix is correct.** `uv run .../sam_schema/cli.py` still sets `__name__ == "__main__"`, so direct execution keeps its protection, while an in-process importer is no longer re-executed. `.pop("PYTHONPATH", None)` is harmless hardening, since the `if` guarantees the key is present.

**The defect and the fix were both reproduced, not read.** Running the PR's own nested-pytest command against unfixed code produced zero bytes of output and exit 0 — the hijack. So the new test is a genuine regression test, and the assertion carrying it is `"1 passed" in result.stdout`. The `returncode == 0` assertion above it is inert, because the hijacked run also exits 0.

### Findings, all being fixed on this branch

**`tests/test_frontend_parity.py:222` — medium — the regression test passes vacuously.** The child environment is `{**os.environ, "PYTHONPATH": "/tmp"}`, which forwards the guard's own sentinel `DH_CLI_PYTHONPATH_CLEARED` whenever the outer process carries it. Concretely: a developer runs `PYTHONPATH=/tmp uv run pytest plugins/development-harness/tests/` against unfixed code, the session imports `sam_schema.cli`, the guard re-executes the whole pytest process with the sentinel set, and this test then inherits it and reports green on the exact defect it guards. Confirmed directly — the same command with `DH_CLI_PYTHONPATH_CLEARED=1` prints `1 passed` against unfixed code, versus empty output without it.

**`sam_schema/cli.py:37` — low, pre-existing — the guard never protects `scripts/run_sam_cli.py`.** That wrapper runs `from sam_schema.cli import app`, and `sam_schema/__init__.py:30` imports `sam_schema.core.dependencies` to `core.models` to `pydantic`. A foreign dependency on `PYTHONPATH` therefore raises inside the package `__init__`, before `cli.py`'s module body and so before the guard. Confirmed with a foreign `pydantic` on `PYTHONPATH`: `uv run --script scripts/run_sam_cli.py plan list` fails with `ImportError: cannot import name 'AliasChoices'` identically with and without the sentinel. This matters because `task_status_hook.py:394` invokes exactly that wrapper.

**A third defect, from CI rather than the review.** "Local / Manifest sync" fails with `The following changed native manifests did not bump their version`. The rebase onto `main` dropped the branch's bump, so its three manifests equal `main`'s. The versioner fixes it.

Nothing else surfaced. The new test inherits `@pytest.mark.integration` from `TestCLIForeignCWD` and so runs in the `test-integration` job, and `run_cli_subprocess`'s `sanitize_env` default does not strip `PYTHONPATH`, so the contamination reaches the child as intended.

PR #3610, opened 2026-09-16
fix(dh): make MCP match the CLI on ledger refusals, and split the SAM server

Jamie-BitFlight:
Follows the work that merged as `aee7ce482`. These commits were made after that merge and are rebased onto current `main`.

## Changes

- Close portability-test gaps for agent and skill path references.
- Add `dh:dh-cli-usage` as the single cross-harness resolver for the grouped DH CLI and server scripts, migrate every caller, and relocate its command and MCP troubleshooting references.
- Centralize ledger refusals so MCP and CLI behavior match for unwritable fields, missing reasons, and ledger-open failures.
- Split `sam_schema/server.py` into focused routing and operation modules.
- Preserve structured GraphQL comment identifiers, fail closed when open-PR searches cannot establish safety, and retain cached-view warning contracts.
- Refresh the Codex skill activation matrix and DH workflow graph.

## Verification

- Required GitHub Actions quality gate: passed.
- Python fast suite: 6,315 passed, 54 skipped, 10 xfailed.
- Cross-backend suites: 60 passed and 24 skipped for each configured matrix lane.
- Integration suite: 19 passed.
- Ruff, ty, markdownlint, plugin validation, manifest sync, file hygiene, Biome, and shell checks: passed.
- Source-backed FastMCP `profile_list` protocol call: succeeded with 77 agent profiles.

Refs #3449


Jamie-BitFlight:
## Review record for this PR

Two no-context reviews ran against this branch before it merged. Both are recorded here, because the second one changed the merged content.

### Review 1 — the CLI migration commits, pinned at `f41a04c12`

Verdict: **CHANGES-REQUIRED**, three findings that held.

| Severity | Finding |
|---|---|
| BLOCKER | `skills/dh-cli-usage/SKILL.md` dropped the clause covering a dh command written without the `<sam_cli/>` token. About 23 shipped files write bare commands — `work-backlog-item/references/workflows/groom/groom-drift.md:38`, `dispatch/SKILL.md:119,152,168,178`, `final-verification/SKILL.md:106` among them. No static test can see that form, so the skill body was the only place it was covered. |
| MAJOR | The `<sam_cli>` and `<dh_scripts>` blocks carried a fourth line naming `$DH_SKILL_DIR`. Nothing in the repository set that variable, and a Bash tool call does not carry an environment variable into the next call, so the command could run with an empty path. |
| MAJOR | The `plan --help` probe became a pre-flight gate on every state-changing command. The design runs it only after a command fails, and gives the reason: a command failing for its own reason (`leased`, `not-ready`) passes `plan --help`, so the caller's own handling applies. |

Clean in the same review: 21 portability tests passing, 27 of 27 CLI-using agents preloading the skill and holding `Bash`, both reference files moved, no dangling references to the old guide, and no agent file gaining a relative path.

### Review 2 — the correction, pinned at `71086b594`

Verdict: **APPROVE**, zero blocking findings.

The body now matches the design's section 3, with one sanctioned deviation: it names Codex, OpenCode and Cursor in the fallback sentence, because the committed test asserts all three names appear. 36 of 40 lines are byte-identical to the design.

- Every command form in the repository resolves: 35 files use `<sam_cli/>`, two use `<dh_scripts/>/name.py`, and about 23 write bare commands. Each has a sentence covering it.
- `git grep DH_SKILL_DIR` returns nothing.
- `tests/test_agent_portability_and_verdict_drift.py`: 21 passed.
- `ruff check` and `markdownlint-cli2`: exit 0.

### One process note

The first reviewer ran two checks against `FETCH_HEAD`, which a concurrent fetch from another session overwrote mid-review. That produced one false finding, which the reviewer caught and disproved by re-running every check in a worktree pinned to the literal commit SHA. Every result above comes from SHA-pinned runs.

PR #3611, opened 2026-09-16
fix(dh): stop refusals escaping the MCP tool wrappers

Jamie-BitFlight:
Every commit here closes one hole: a refusal that raises a type no MCP tool wrapper catches, so the tool call fails with an unhandled exception instead of the documented `error` field.

## How this PR reached fifteen commits

It opened as five hygiene fixes for a red `main`. #3610 merged as `33fe7daea` and carried four of them, so the branch was reset onto `main` with the one defect that remained. Each review round then widened the search and found more of the same defect: two sites in the diff, three more in the module, seven across `gh_client`, five in the package, one inside a function already fixed, and finally a whole class the earlier sweep had wrongly called safe.

One commit was dropped rather than rebased. It removed a `backend unreachable — ` prefix from a test assertion, and `main` emits that prefix at `operations.py:3362` while asserting it at `test_refusal_not_item_missing.py:160`.

## The defect

A tool wrapper such as `backlog_update` (`server.py:2747`) catches `BacklogError` and nothing else. `ValidationError` is the `BacklogError` subclass every other validation failure in `backlog_core` already uses.

The last round found that converting a `raise ValueError` is not always sufficient. A `ValueError` raised inside a pydantic validator comes back out as `pydantic.ValidationError`, which subclasses `ValueError` and is not a `BacklogError`, so it escapes exactly like the bare form. `artifact_get(item_id="")` escaped all four `artifact_*` tools this way.

## What each group of commits does

**The two issue mirrors.** `_rename_item_title` and `_apply_plan_to_item` raised `ValueError` for a non-numeric issue reference, and their handler caught `(GithubException, BacklogError)` only. Both now raise `ValidationError`, which that handler catches, so a caller gets `True` plus a warning — correct, because both functions write to the backend before the GitHub branch, and the docstring already promised the backend update survives an unavailable GitHub. The same handlers also missed transport errors, since PyGithub wraps no `requests` exceptions; both now spread `*RETRYABLE_TRANSIENT_EXCEPTIONS`, as three siblings in the file already do.

**Seven sites in `gh_client`.** All raised the same refusal on the same parse, in three shapes: four sit inside a handler and now warn, three sit in functions with no handler and now propagate as `BacklogError`, which the owning wrapper turns into its `error` response.

**Two label-name sites.** Label names read off a live GitHub issue meet a pattern that rejects emoji and any `/`-namespaced name such as `area/api`.

**The backend name and the entry ids.** `create_backend` refused an unknown backend name; `_parse_entry_timestamp` refused an id with no ISO prefix, and then refused a calendar-impossible date one line later. `_ISO_TIMESTAMP_RE` checks shape, and `datetime.fromisoformat` checks whether the date exists, so `2026-13-01` passed the first and failed the second.

**The artifact tools.** `_manifest_reference` is the one function all four `artifact_*` tools use to turn caller input into a `ContentRef`, so the conversion sits there. It converts every validator refusal that reaches it, of which exactly one branch is reachable from caller input today: `models.py:1342`, on an empty `item_id`. `artifact_content_reference` in `artifact_manifest_store.py` needed the same treatment separately, because it builds its reference without passing through `_manifest_reference`, and both of its callers are covered by that one guard.

The 13 validators themselves stay as they are, and that is deliberate. A `BacklogError` raised inside a pydantic validator propagates unwrapped, so converting them would bypass `file_cache_state._read:298`, `_salvage_field:785` and `parse_work_item_head:136` — killing corrupt-cache salvage and the fail-closed path.

**The SAM context backend refuses two ways.** `context_config.py` raises `SamError` for an unrecognised name and `NotImplementedError` for `"github"`, which is a recognised name listed in its own "Valid options" message. The first guard caught only the former, so a user following that message met a raw `NotImplementedError`. The guard now catches both, matching the CLI sibling it documents itself as mirroring.

**The SAM server.** `task_config.py` and `context_config.py` refused an unknown backend name with a bare `ValueError`, and `server_backend.py:25` guarded only `RuntimeError`. `CONTEXTBACKEND=bogus` therefore made `import sam_schema.server_backend` fail outright, so the server never started. The bootstrap now runs inside `get_context_backend()`, the refusal is `SamError`, and the boundary converts it to `ToolError`. The server starts, unrelated tools keep working, and only the tool needing a context backend fails, naming the bad value.

**Three documentation corrections.** A docstring claimed `backlog_update` reports a refusal it never sees. Another justified a conversion by a caller that does not exist. A comment gave a false reason for why an empty status map is harmless. `create_task_issue`'s third outcome — the issue created with no labels — is now written where its callers read it.

## Checks

Each behavioural commit reproduced its defect first, through the path a caller uses, and recorded the failure before the change.

From `plugins/development-harness`: `uv run pytest backlog_core/tests tests_backlog tests tests_sam` gives 5999 passed, 54 skipped, 133 deselected, 6 xfailed.

From the repository root: `uv run ruff check`, `uv run ruff format --check` and `uv run ty check` each exit 0.

`CONTEXTBACKEND=bogus uv run python -c "import sam_schema.server_backend"` exits 0.

## What this PR does not claim

An earlier sweep classified the remaining `raise ValueError` sites in `backlog_core` and put 13 of them in a category called "fenced by a pydantic validator". That category is wrong: `pydantic.ValidationError` subclasses `ValueError` and is not a `BacklogError`, so a refusal raised inside a validator escapes a tool wrapper exactly like a bare one. `artifact_get(item_id="")` proved it, and that site is fixed here.

So this PR does not claim the defect class is closed. What it claims is narrower and checked:

- Every site it converts is reproduced first through the path a caller uses, and its `Raises:` text states what a caller observes.
- The three `raise ValueError` sites in the SAM server modules are `case _:` fallbacks, fenced by the discriminated-union `Literal` that pydantic validates at the tool boundary before the body runs.
- The validator sites in `models.py` and `entry_blocks.py:509` are now classified, each with the guard that fences it named and pinned by a test. One of the eight was reachable, and it is fixed here.

That one is `models.py:1005`, `BacklogItem._validate_added`, reached from the `backlog_pull` tool: the tool reads a remote issue body, and its free-form metadata block feeds `BacklogItem(added=…)`. The failure was worse than a crash — FastMCP reads a stray `pydantic.ValidationError` as an input-schema failure, so the caller was told its own arguments were invalid when the fault was a malformed remote issue body. The conversion sits at the parse boundary in `github_sync.parse_issue_body`, because `file_cache_state._salvage_field`, `FileCache._work_item_snapshots` and the beads handler all depend on the validator still raising a `ValueError` subclass to salvage or dead-letter a corrupt cached item.

The 13 validators stay unconverted for a reason that outlives this PR: a `BacklogError` raised inside a pydantic validator propagates unwrapped, so converting them bypasses `file_cache_state._read`, `_salvage_field` and `parse_work_item_head`, which catch `pydantic.ValidationError` to salvage a corrupt cache and to fail closed. A reachable validator therefore gets converted at the boundary where its model is built from caller input, as `_manifest_reference` does, never inside the validator.

Refs #3449


Jamie-BitFlight:
## `/code-review high` on `093e4308f`

Scope: `git diff origin/main...HEAD` — three manifest bumps, the two `ValueError` to `ValidationError` changes with their docstrings, one mock added to an existing test, and a new 78-line test file.

**The core fix is correct.** `ValidationError` subclasses `BacklogError` at `models.py:702`, so both refusals are now caught by their enclosing handler instead of escaping. `git grep "raise ValueError"` in `operations.py` returns nothing.

**The second commit is justified, checked empirically.** The pre-fix copy of that test fails with `NetworkBlocked: Blocked DNS resolution of 'api.github.com'` through `batch_fetch_statuses` to `try_get_github` to `gh.get_repo`. `gh_client.batch_fetch_statuses` resolves `try_get_github` as a module global, so the patch point is the right seam. The empty status map weakens nothing: `_filter_open_items` reads it only under the `status` filter at `operations.py:1945`, and this call passes none.

### Two findings, both being fixed on this branch

**`operations.py:806` — transport errors still escape.** The handler catches `(GithubException, BacklogError)`, and its twin at `:890` does the same. Neither covers `requests.exceptions.ConnectionError` or `Timeout`: `_graphql_request` converts only `GithubException`, and PyGithub wraps no `requests` exceptions. A network failure during `backlog_update(title=...)` therefore still fails the tool call with an unhandled exception — the exact failure this PR closes — after the backend write has landed. Three sibling handlers in the same file already spread `*RETRYABLE_TRANSIENT_EXCEPTIONS`: `:263`, `:1563` and `:2157`.

**`operations.py:770` — the docstring misstates what the wrapper sees.** Its `Raises:` section ends "which is the documented surface `backlog_update` (`except BacklogError`) reports". `backlog_update` never sees the refusal, because the handler catches it inside the function. The tool returns its normal success shape with the refusal only in `warnings`.

Nothing else surfaced. No caller of either function depended on `ValueError` propagating; no test asserted the old behaviour; warnings reach the MCP wire, because `ToolResponse` inherits `messages`, `warnings` and `errors` from `Output`; and the version bump is self-consistent.

Issue #2437, opened 2026-05-24
refactor: add-new-feature orchestrator must not read or inject artifact content between phases


Jamie-BitFlight:
## Description

Phase 3 of add-new-feature routes through dh:task-worker with a specialist profile loaded via profile_load instead of dispatching directly to a dedicated architect agent. This causes the orchestrator to read Phase 1 and Phase 2 agent outputs inline into its own context window, process them, and inject conclusions into the Phase 3 prompt as guiding facts — analysis that belongs to the architect, not the orchestrator.

In a live session this caused: wrong path arithmetic injected as stated facts, a fabricated open question about adding rich to PEP 723 dependencies (a project policy violation per .claude/rules/language-conventions.md), and errors the architect had to self-correct. The task-worker intermediary also loses the specialist agent's own methodology for verifying claims against primary sources before writing a spec.

Fix: dispatch Phase 3 directly to the specialist architect agent with only artifact IDs (Phase 1 feature-context, Phase 2 codebase-analysis), the issue selector, and domain skills. The architect reads those artifacts itself, checks claims against project rules, and produces the spec. The orchestrator never reads artifact content between phases — it only tracks that artifacts exist via artifact_list and passes their identifiers forward.

## Fact-Check

<div><sub>0000-00-00T00:00:00Z</sub>

CLAIM: Phase 1 and Phase 2 agents return their full findings as inline text to the orchestrator
VERDICT: VERIFIED
EVIDENCE: SKILL.md line 173 explicitly documents Phase 2: "The agent returns the content in its response; the orchestrator then registers it as an artifact." Phase 1 template (lines 142-147) similarly shows artifact registration metadata but not a content-passing mechanism, implying orchestrator reads inline.
SOURCE: plugins/development-harness/skills/add-new-feature/SKILL.md:173, lines 142-147 (accessed 2026-05-24)

CLAIM: Phase 3 already has the correct contract — orchestrator receives only STATUS: DONE (fixed in commit 9ef49696, issue #2429)
VERDICT: VERIFIED
[...]

PR #3617, opened 2026-09-17
fix(dh): restore portable CLI handoffs

Jamie-BitFlight:
## Summary

Rebuilds only the unique portability and documentation tranche from #3593 on current main:

- cross-harness DH CLI root resolution with verified Cursor MCP fallback and fail-closed filesystem behavior
- CLI preload and token portability checks
- MCP recovery documentation
- workflow-map reference relocation
- no-issue-ID final-handoff handling

The Python-version, GraphQL, TLS, certificate, and status changes from #3593 were not copied because #3551 merged stronger implementations.

## Verification

- portability suite: 27 passed with warnings treated as errors
- development-harness suite: 6499 passed, 54 skipped, 6 xfailed
- full suite: 6579 passed, 54 skipped, 10 xfailed
- changed-file hooks, skilllint, compatibility annotations, and diff checks passed

Replaces the unique portion of #3593.

<h2><a href="https://app.greptile.com/api/retrigger?id=65402493"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a></h2>

The PR appears safe to merge, with a non-blocking generated-artifact consistency issue that should be corrected.

<sub>Reviews (3) · Last reviewed commit: ["fix(dh): address portability review find..."](https://github.com/jamie-bitflight/claude_skills/commit/85773b0f8d7f426b0846a731406de5b73d0ab7e8)</sub>


plugins/development-harness/CLAIMS-REGISTER.md
```diff
@@ -89,6 +89,13 @@ Consequences the design draws, each a claim in its own right:
   Cursor repositories and the plugin template's validator, and did not find it. Confidence:
   source, Cursor by snippet. Re-check: section 4 of each file.
   `plugins/plugin-creator/CLAIMS-REGISTER.md` holds this claim for the skill-portability rule.
+- **Cursor does not have an established absolute skill-root exposure mechanism.** Its skills
```
Jamie-BitFlight: this should be researched online definitely. not over explained either no evidence

plugins/development-harness/docs/work-ledger/measurements/harness-cursor.md
```diff
@@ -247,6 +247,12 @@ optional `scripts/`, `references/`, `assets/`. Frontmatter: `name`, `description
 your SKILL.md using relative paths from the skill root." Plugin skills: `<plugin>/skills/<name>/SKILL.md`
 `[repo + docs-snippet]`.
 
+The skills documentation establishes resolution of bundled-resource references relative to the skill root. It
```
Jamie-BitFlight, line 250: what does a skill root reference in a skill have to do with a plugin's mcp configuration?

plugins/development-harness/skills/complete-implementation/references/final-handoff.md
```diff
@@ -9,7 +9,9 @@ The following diagram is the authoritative procedure for Final Handoff Output. E
 
 ## Concerns Check — Implementation Notes
 
-The first step reads the Concerns section of the just-resolved issue before routing to the next work item.
+When a GitHub issue number exists, the first step reads its Concerns section before routing to the
```
Jamie-BitFlight: why are you saying that the universal concerns section is github specific? it's a way for passing concerns found during a subagent task back up stream s d to the orchestrator and also store it in the backend for reference

plugins/development-harness/skills/dh-cli-usage/SKILL.md
```diff
@@ -1,40 +1,60 @@
...
+description: Use when a development-harness agent or skill needs to run the DH CLI or locate DH server scripts across supported harnesses.
 user-invocable: false
 ---
 
-# dh CLI usage
+# DH CLI Usage
+
+Determine this skill's absolute directory with the branch for the current harness:
+
+- Claude Code: use `${CLAUDE_PLUGIN_ROOT}/skills/dh-cli-usage`. Claude Code substitutes
```
Jamie-BitFlight: this whole paragraph needs correction, why was it even added? what was wrong with the system that this much explanation was required?
how was this evaluated?

plugins/development-harness/skills/dh-cli-usage/references/mcp-connection-check.md
```diff
@@ -1,34 +1,73 @@
...
+- When tool search is disabled, Claude Code uses the `WaitForMcpServers` tool to
+  wait for connecting servers before proceeding.
+
+Apply the troubleshooting steps below only when a server has genuinely failed to
+connect — i.e. `/mcp` shows the server as **failed**, or a tool call returns a
+connection error (not a transient "still connecting" state).
+
+## Troubleshooting a Failed Server
+
+If a server shows as failed in `/mcp` or tool calls return connection errors:
```
Jamie-BitFlight: why is this written to a human? there's no human that reads the skill at runtime

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
```diff
@@ -177,12 +177,12 @@
 # skill, never re-derive either value itself. These patterns are the three ways a file did that
 # derivation directly instead: a raw ``${...PLUGIN_ROOT}`` template variable, a
 # ``${...SKILL_DIR}/..`` parent-directory climb written out longhand, and the CLI's literal path.
-# Only Claude Code substitutes a plugin-root or skill-dir template variable in a skill or agent
-# body; every other measured harness ships one to the model as raw text, so a reference or doc
-# file carrying it ships broken text everywhere else (``CLAIMS-REGISTER.md``'s
-# ``${CLAUDE_PLUGIN_ROOT}`` entry; ``rules/runtime-vs-design-time.md``). ``dh-cli-usage`` instead
-# derives the root from the skill's own directory, which every measured harness except Cursor gives
-# the model.
+# Claude Code substitutes its plugin-root variable, Kimi and Hermes substitute their skill-root
```
Jamie-BitFlight: this paragraph needs fact checking and rewriting to be short, or removed.

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
```diff
@@ -243,12 +243,14 @@
 DH_CLI_USAGE = DH_CLI_USAGE_DIR / "SKILL.md"
 DH_CLI_USAGE_SKILL_URI = "dh:dh-cli-usage"
 
-# The three forms a harness substitutes a skill's own directory into its body as.
-# ``dh-cli-usage`` derives the plugin root and the CLI's location from one of these -- never from a
-# plugin-root variable, which only Claude Code resolves.
-SKILL_DIR_VARIABLES = ("CLAUDE_SKILL_DIR", "KIMI_SKILL_DIR", "HERMES_SKILL_DIR")
-SAM_CLI_LINES = tuple(f'uv run "${{{v}}}/../../sam_schema/cli.py"' for v in SKILL_DIR_VARIABLES)
-DH_SCRIPTS_LINES = tuple(f"${{{v}}}/../../scripts" for v in SKILL_DIR_VARIABLES)
+FIRST_CLASS_HARNESSES = ("Claude Code", "Codex", "OpenCode", "Cursor")
```
Jamie-BitFlight: what is this trying to achieve and have you proved it works

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
```diff
@@ -965,6 +977,35 @@ def test_cli_guide_and_connection_check_live_in_dh_cli_usage() -> None:
...
+    coverage = refresh_dir / "references" / "COVERAGE.md"
+    methodology = refresh_dir / "references" / "workflow-trace-methodology.md"
+    rule = PLUGIN_ROOT / ".claude" / "rules" / "workflow-extraction.md"
+
+    assert coverage.is_file()
+    assert methodology.is_file()
+    assert not (PLUGIN_ROOT / "docs" / "workflow-layers" / "COVERAGE.md").exists()
+    assert not (PLUGIN_ROOT / "docs" / "workflow-trace-methodology.md").exists()
+
+    active_text = "\n".join(path.read_text(encoding="utf-8") for path in (refresh_dir / "SKILL.md", methodology, rule))
```
Jamie-BitFlight, line 1067: does this make the product better? does it justify its own maintenance burden?

Jamie-BitFlight (commented):
this needs serious checking and evaluation of the claims made, and adversarial review of the content, both new and old in the files edited.
The claims about how some functionality works in different harnesses needs verification

plugins/development-harness/skills/meta-workflow-graph-refresh/references/COVERAGE.md
Jamie-BitFlight, line 1: **Broken relative links after the move (content was renamed +0/-0, so its own `../` links were never re-based).**

`COVERAGE.md` moved from `docs/workflow-layers/` to `skills/meta-workflow-graph-refresh/references/`, but three in-file links still assume the old depth:

- L104 and L131: `` `../workflow-trace-methodology.md` `` — used to resolve to `docs/workflow-trace-methodology.md`. It now resolves to `skills/meta-workflow-graph-refresh/workflow-trace-methodology.md`, which does not exist. The file is now a **sibling**, so this must be `./workflow-trace-methodology.md`.
- L104 and L135: `` `../plans/prd-workflow-extractor.md` `` — used to resolve to `docs/plans/prd-workflow-extractor.md`. It now resolves to `skills/meta-workflow-graph-refresh/plans/prd-workflow-extractor.md`, which does not exist. The real file is still at `plugins/development-harness/docs/plans/prd-workflow-extractor.md` (verified present), i.e. `../../../docs/plans/prd-workflow-extractor.md` from here.

Concrete failure: an agent follows the rule file (`.claude/rules/workflow-extraction.md` L48) to read `references/COVERAGE.md`, reaches Tier-2 step 2, and is told to open `../workflow-trace-methodology.md` for "the extraction protocol and field definitions per layer type". The read fails, and the extraction protocol is exactly the content this PR moved to keep reachable.

Note `workflow-trace-methodology.md` in this same PR *did* get its sibling link fixed (`[COVERAGE.md](./COVERAGE.md)`), so the rebasing pass just missed this file.

---

plugins/development-harness/skills/meta-workflow-graph-refresh/references/workflow-trace-methodology.md
Jamie-BitFlight, line 1: **L130 still carries the pre-move relative path.**

This file's link to `COVERAGE.md` was correctly re-based to `./COVERAGE.md` on L7 in this PR, but L130 was left:

```
retired; the replacement pipeline is pending redesign (see the PRD at
`../plans/prd-workflow-extractor.md`). The 4-phase ensemble it ran internally:
```

From `skills/meta-workflow-graph-refresh/references/` that now resolves to `skills/meta-workflow-graph-refresh/plans/prd-workflow-extractor.md`. The file is still at `plugins/development-harness/docs/plans/prd-workflow-extractor.md`, so the correct relative path is `../../../docs/plans/prd-workflow-extractor.md`.

Failure scenario: an agent running the extraction protocol hits the "pipeline is pending redesign" note, opens the cited PRD to find out what replaces the retired `dh:workflow-extractor`, gets a not-found, and either stops or improvises the extraction pass.

---

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
```diff
@@ -965,6 +1052,73 @@ def test_cli_guide_and_connection_check_live_in_dh_cli_usage() -> None:
...
```
Jamie-BitFlight, line 1067: **The guard scans every relocated file except the one that actually has stale links.**

`active_text` covers `SKILL.md`, `workflow-trace-methodology.md` and the rule file, but not `COVERAGE.md` — and `COVERAGE.md` is the file this PR renamed with zero content edits, so it is the one still carrying pre-move `../` paths (L104, L131 → `../workflow-trace-methodology.md`; L104, L135 → `../plans/prd-workflow-extractor.md`).

The test passes today while both documents the skill points a reader at are unreachable from `COVERAGE.md`. Adding `coverage` to the tuple would fail right now, which is the point of the guard.

The two `not ... .exists()` assertions above also only prove the old copies were deleted; nothing checks that what the moved files *point at* still resolves. A stronger shape: resolve every relative path written in the two relocated files against their new directory and assert the target exists.

---

plugins/development-harness/skills/complete-implementation/references/final-handoff.md
```diff
@@ -9,22 +9,24 @@ The following diagram is the authoritative procedure for Final Handoff Output. E
 
 ## Concerns Check — Implementation Notes
 
-The first step reads the Concerns section of the just-resolved issue before routing to the next work item.
+The first step reads the Concerns section of the just-resolved item before routing to the next work
+item. When the item carries no backend reference, skip that read and go to the slug search; a
+missing reference is not a backend failure.
 
-**Selector source (`#{N}`)**: The resolved issue number is available from the **"Apply status:verified Label" Step 2** of the `complete-implementation` skill — that step calls `backlog_list(title="{slug}")`, finds the matched item, and extracts `issue_number` from the item's `issue` field (e.g., `"#2437"` → `issue_number = 2437`). Use that integer as `#{N}` in the `backlog_view` call below.
+**Selector source (`{item_ref}`)**: `{item_ref}` is what the `complete-implementation` skill stored in **Step 3 -- Extract context for proportional gates**: the `backlog_view` response's opaque `reference`. That step runs on either backend. The value is a GitHub integer on a GitHub backend and a string such as `bd-a3f8` on beads; pass it unchanged as `{item_ref}` in the `backlog_view` call below.
```
Jamie-BitFlight, line 16: **The new pointer is backend-neutral but not entry-path-neutral — it names a step the plan-address path explicitly skips.**

`complete-implementation/SKILL.md` L57-58 opens *Resolve Issue* with: "Entered when input is `#N`, bare `N`, GitHub URL, or another work-item reference such as `bd-a3f8`. … **Skip for plan address input.**" Step 3 (L93) lives inside that section, so on the plan-address entry path it never runs. That path sets `{item_ref}` somewhere else entirely — *Resolve Plan Address* (L262): "use the plan row's `feature` field as `{slug}` and its `issue` field as `{item_ref}` when present."

Both paths reach this file (SKILL.md L687 is the single Final Handoff entry point), so the fix swapped one incomplete pointer for another: the old text named a section beads skips, the new text names a step the plan path skips. Concretely — `/dh:complete-implementation P12` on a GitHub backend with a plan whose `issue` field is set: `{item_ref}` is populated, the handoff correctly calls `backlog_view`, but the stated source for that value is a step that did not execute, so anyone (or any agent) auditing the handoff against SKILL.md finds no such stored value and cannot tell whether the selector is trustworthy.

There is a third source too: *Apply status:verified Label* (L600-601) recovers `{item_ref}` by slug lookup when the plan exposed no owner reference. `test_final_handoff_takes_the_item_reference_from_a_backend_neutral_step` now asserts `"Apply status:verified" not in final_handoff`, which blocks documenting that source here at all.

Suggest naming the value rather than one producing step — e.g. "`{item_ref}` as resolved earlier in this run (Step 3 on the work-item path, the plan row's `issue` field on the plan-address path, or the slug fallback)" — and relaxing the test to assert the handoff does not point at a *single* conditionally-skipped step, rather than banning one section name.

---

plugins/development-harness/skills/dh-cli-usage/references/mcp-connection-check.md
```diff
@@ -1,34 +1,34 @@
...
-   ```text
+1. Make the failed call once more. A server that was still starting answers on the second call.
+
+2. When it fails again, start each server directly and read what it prints:
+
+   ```bash
+   uv run --script "<dh_scripts/>/run_backlog_server.py"
    uv run --script "<dh_scripts/>/run_sam_server.py"
-   uv run --script "<dh_scripts/>/run_backlog_server.py" --project-dir .
    ```
```
Jamie-BitFlight, line 14: **Two long-running stdio servers presented as one unbounded, foreground `bash` block.**

`run_backlog_server.py` and `run_sam_server.py` both end in `mcp.run()` (stdio transport), so each blocks until stdin closes. The prose after the block says "A process that stays up and prints no error is a working server; stop it before you start the next one" — but the recovery step hands the model a single fenced block with both commands in it, and the natural action is to execute the block.

Failure scenario: an agent recovering from a failed `mcp__plugin_dh_*` call runs the block on a harness whose shell keeps a tty or an open stdin. Command 1 never returns, the second command never runs, and the session that was already blocked on a broken MCP server is now blocked on a hung Bash call with no stated timeout and no stated way out.

`AGENTS.md` "Gotchas" #4 is explicit about this case: wrap any external command that may hang with `uv run --script scripts/run_bounded.py --timeout-seconds <n> -- <command>`. Two separate blocks, each bounded, would keep the diagnostic value and make "it stayed up" an observable outcome (timeout expiry) instead of a hang.

Separately: the old step passed `--project-dir .` to `run_backlog_server.py` and this drops it. That matches `.mcp.json` (which sets `cwd: "."` and passes no flag), so the behaviour is consistent — but it now means the check only reproduces the real server's project binding when it is run from the repo root, and nothing here says so.

---

Jamie-BitFlight (commented):
Reviewed the diff excluding `graphify-out/` (generated). Ran `uv sync` and the changed test module: **29 passed**. Re-ran `scripts/generate_harness_compatibility.py` — the committed blocker counts match the generator's output exactly, so `harness_compatibility.json` is accurate. `skilllint` passes on the touched skill files.

5 findings, most-severe first. The cluster is one theme: the `meta-workflow-graph-refresh` document move re-based the *inbound* pointers but not the *outbound* ones, and the new guard added alongside it scans every relocated file except the one that broke.

1. `references/COVERAGE.md` — 4 broken `../` links after the move (renamed +0/-0, content never re-based).
2. `references/workflow-trace-methodology.md` L130 — `../plans/prd-workflow-extractor.md` left behind when L7 was fixed.
3. `tests/…verdict_drift.py` L1067 — `active_text` omits `COVERAGE.md`, so the guard passes over the exact defect it exists to catch.
4. `final-handoff.md` L16 — `{item_ref}` sourced from Step 3, which `SKILL.md` L58 tells the plan-address entry path to skip. Backend-neutral, but not entry-path-neutral.
5. `mcp-connection-check.md` L11-14 — two blocking stdio servers in one unbounded foreground block, against `AGENTS.md` gotcha #4.

Findings 1-3 are a single fix plus a guard that would then catch it. Nothing here blocks the portability work itself, which reads well — the `<sam_cli>`/`<dh_scripts>` one-complete-line-per-variable shape and the fail-closed path are a clear improvement over hand-assembling a path, and `CLAIMS-REGISTER.md` warrants each source properly.

---

plugins/development-harness/skills/complete-implementation/references/final-handoff.md
Jamie-BitFlight, line 16: **`backlog_view` does not return a `reference` key — this makes the Concerns check dead on every path.**

This rewrites the `{item_ref}` source to "the `backlog_view` response's opaque `reference`". That field does not exist in the response:

- `ViewItemResult` (`backlog_core/models.py:1612`) declares `title, priority, description, source, added, plan, issue, file_path, groomed, status, number, state, body, labels, milestone, ...` — no `reference`.
- `BacklogViewResponse` (`backlog_core/tool_responses.py:932`) likewise has `issue: str | None` and `number: int | None`, no `reference`.
- `view_result_from_local_item` (`backlog_core/parsing.py:1388`) populates `issue=item.issue` and never a `reference`.

`BacklogItem.reference` exists internally, but it is `issue or sha256(title)` (`models.py:1009 _derive_stable_reference`) and is not serialised into either view response.

Scenario: complete-implementation finishes item `#2437`. The agent reaches Final Handoff, looks for `reference` in the stored Step 3 context, finds nothing, and takes the new `HasRef -->|"No"| Fetch` branch added in this same diff. The Concerns read is skipped silently — and by design of the new error-handling paragraph, no warning is emitted either. Unresolved concerns are never surfaced, on GitHub and on beads alike.

The backend-neutral field that does exist is `issue`: `"#42"` on GitHub, `"bd-a3f8"` on beads. Point `{item_ref}` at that (and fix the matching line in `complete-implementation/SKILL.md:97`).

---

plugins/development-harness/skills/dh-cli-usage/SKILL.md
```diff
@@ -18,13 +18,17 @@ ${KIMI_SKILL_DIR}/../../scripts
...
-absolute. A line still reading `${…}` names a variable this harness does not fill in; pass over it.
+Use the line whose path is absolute. A line still reading `${…}` names a variable this harness does
+not fill in; pass over it.
 
-Codex, OpenCode and Cursor fill in no line above. When no line holds an absolute path, take the
-directory your harness stated when it loaded this skill: a `Base directory for this skill:` line, a
-`References are relative to` line, or the directory of the `SKILL.md` you read. Then `<sam_cli/>` is
-`uv run "<that directory>/../../sam_schema/cli.py"` and `<dh_scripts/>` is `<that directory>/../../scripts`.
+When no line is absolute, take the directory your harness stated for this skill: a
+`Base directory for this skill:` line above this body, or the `skill_root` value a `skills.read`
```
Jamie-BitFlight, line 25: **"above this body" contradicts this PR's own CLAIMS-REGISTER entry for the harnesses that actually need this fallback.**

The register entry added in this same diff states:

> OpenCode and Kilo Code **append** `Base directory for this skill: <dirname of SKILL.md>` **after the body** (`packages/opencode/src/tool/skill.ts`, read 2026-09-06) … and Claude Code emits the same line (measured 2026-09-18).

Claude Code is the only one that emits it *before* the body, and Claude Code substitutes `${CLAUDE_SKILL_DIR}` so it never reaches this paragraph at all. The harnesses that do reach it — OpenCode and Kilo Code — put that line *below* the body.

Scenario: a dh skill loads `dh:dh-cli-usage` under OpenCode. No `<sam_cli>` line is absolute. The model follows this instruction, looks only above the body, finds no `Base directory for this skill:` line, and falls through to the next paragraph's `STATUS: BLOCKED`. The dh CLI is unreachable on OpenCode and Kilo Code even though the directory was stated.

Drop the positional qualifier (the previous wording, "the directory your harness stated when it loaded this skill", had no such constraint).

---

plugins/development-harness/skills/dh-cli-usage/references/mcp-connection-check.md
```diff
@@ -1,34 +1,34 @@
...
 
-1. Inspect the harness's MCP server status. If the DH backlog and SAM servers are connected, rerun
-   the original tool call.
-2. Restart the agent session so plugin MCP servers restart.
-3. If the failure persists, load `dh:dh-cli-usage` and run both source commands below, substituting
-   `<dh_scripts/>` from that skill:
+You are here because an `mcp__plugin_dh_backlog__*` or `mcp__plugin_dh_sam__*` call failed.
 
-   ```text
+1. Make the failed call once more. A server that was still starting answers on the second call.
```
Jamie-BitFlight, line 7: **The rewrite drops the two remediations that actually fix a failed MCP connection.**

Removed from the previous procedure:

- old step 2: "Restart the agent session so plugin MCP servers restart."
- old step 5: "Check the harness's MCP startup timeout and restore its default when a local override aborts startup before either source command initializes."

A too-low `MCP_TIMEOUT` is the single most common cause of "server failed to connect", and the new procedure has no step for it. Scenario: a session sets `MCP_TIMEOUT=5000`; both dh servers abort during startup. Step 1 retries and fails. Step 2 starts each server by hand — both come up fine standing alone, because no startup timeout applies there — so the agent gets no signal. Step 3 routes to the CLI, and step 4 reports `STATUS: BLOCKED` for anything MCP-only. The actual one-line fix is never surfaced.

The CLAIMS-REGISTER entry added here notes the old citation "covered Claude Code's `MCP_TIMEOUT` and connection-waiting behaviour only" — that is a reason to scope the step to Claude Code, not to delete it.

---

plugins/development-harness/skills/dh-cli-usage/SKILL.md
```diff
@@ -18,13 +18,17 @@ ${KIMI_SKILL_DIR}/../../scripts
...
-Your harness filled in one line of each block with an absolute path. Use the line whose path is
-absolute. A line still reading `${…}` names a variable this harness does not fill in; pass over it.
+Use the line whose path is absolute. A line still reading `${…}` names a variable this harness does
+not fill in; pass over it.
 
-Codex, OpenCode and Cursor fill in no line above. When no line holds an absolute path, take the
-directory your harness stated when it loaded this skill: a `Base directory for this skill:` line, a
-`References are relative to` line, or the directory of the `SKILL.md` you read. Then `<sam_cli/>` is
-`uv run "<that directory>/../../sam_schema/cli.py"` and `<dh_scripts/>` is `<that directory>/../../scripts`.
+When no line is absolute, take the directory your harness stated for this skill: a
```
Jamie-BitFlight, line 24: **Dropping "the directory of the `SKILL.md` you read" makes the dh CLI permanently unreachable on Cursor.**

The previous body offered three fallbacks: a `Base directory for this skill:` line, a `References are relative to` line, or *the directory of the `SKILL.md` you read*. Only the first survives (plus the new Codex `skill_root`).

Cursor substitutes no skill-directory variable and (per the `harness-cursor.md` note added in this diff) states no absolute skill root. With the read-path fallback gone, every Cursor run of every dh skill that resolves `<sam_cli/>` now lands on the `STATUS: BLOCKED` paragraph below.

Cursor is a first-class target for this plugin — `.cursor-plugin/plugin.json` is maintained and version-bumped in this same PR. And the register entry this change rests on is explicitly labelled *"Confidence: unestablished — absence of evidence, not evidence of absence."* Failing closed on an unestablished claim removes a path that worked; the model does have the SKILL.md's path whenever it opened the file.

---

plugins/development-harness/.claude/rules/workflow-extraction.md
```diff
@@ -1,7 +1,6 @@
 ---
 paths:
-- docs/workflow-layers/**
-- docs/workflow-trace-methodology.md
+- skills/meta-workflow-graph-refresh/references/**
```
Jamie-BitFlight, line 3: **The new `paths` glob stops this rule firing on the layer JSON files, which did not move.**

`docs/workflow-layers/**` is replaced by `skills/meta-workflow-graph-refresh/references/**`, but only `COVERAGE.md` moved out of `docs/workflow-layers/`. Still there: `L0-forks.json`, `L1-trace-*.json`, `G1`–`G8`, `KNOWN_ENTITIES.md`, `SCOPE.md`, `extraction-rules.json`.

Scenario: someone edits `docs/workflow-layers/G4-concurrency.json`. Nothing under `skills/meta-workflow-graph-refresh/references/` is touched, so this rule never loads — including its own "Do not edit the layer JSON files by hand — extract from source and let the assembler write" instruction and the corroboration-threshold rules that govern what may enter those files. The rule now fires only when someone edits the two prose documents, which is the case that needs it least.

Keep both globs.

---

Jamie-BitFlight (commented):
Code review of the portability/handoff changes. 8 findings, most severe first. The test suite at this head passes (29 passed), and `scripts/generate_harness_compatibility.py --check` reports the table current, so these are behavioural/doc-contract issues the guards do not cover.

---

plugins/development-harness/skills/meta-workflow-graph-refresh/SKILL.md
```diff
@@ -1,12 +1,12 @@
...
-description: Entry point for refreshing the DH workflow graph after changes to skills, agents, Mermaid flowcharts, MCP tools, or artifact flows. Use after any structural workflow change to keep docs/dh-workflow-graph.json current.
+description: Entry point for refreshing the DH workflow graph. Use when skills, agents, Mermaid flowcharts, MCP tools, or artifact flows change and docs/dh-workflow-graph.json must remain current.
 ---
 
 # Workflow Graph Refresh
 
 The update process is fully documented in two files. Read them in order:
 
-1. `../../docs/workflow-layers/COVERAGE.md` — identifies which layers are affected by your change type, the Tier 1 / Tier 2 update process, and what not to do
+1. [references/COVERAGE.md](./references/COVERAGE.md) — identifies which layers are affected by your change type, the Tier 1 / Tier 2 update process, and what not to do
```
Jamie-BitFlight, line 10: **The moved `COVERAGE.md` keeps relative links that were correct at its old depth and are broken at the new one.** (Anchored here — `references/COVERAGE.md` is a pure rename with no diff hunk to comment on.)

At the old path `docs/workflow-layers/COVERAGE.md` these resolved; at `skills/meta-workflow-graph-refresh/references/COVERAGE.md` they do not:

- line 104 and line 131 — `` `../workflow-trace-methodology.md` `` → `skills/meta-workflow-graph-refresh/workflow-trace-methodology.md`. The file is a sibling now; this should be `./workflow-trace-methodology.md`.
- line 104 and line 135 — `` `../plans/prd-workflow-extractor.md` `` → `skills/meta-workflow-graph-refresh/plans/…`. The real path is `plugins/development-harness/docs/plans/prd-workflow-extractor.md`.

Scenario: an agent reads COVERAGE.md as instructed by this SKILL.md, reaches Tier 2 step 2, follows `../workflow-trace-methodology.md`, and gets a missing file — with the correct copy sitting next to the one it just read.

`test_workflow_refresh_references_follow_the_relocated_documents` does not catch this: `active_text` joins `SKILL.md`, `methodology` and the rule file, and deliberately omits `COVERAGE.md`. Adding `coverage` to that tuple would also need the `../` forms asserted against, not just the `docs/…` ones.

---

plugins/development-harness/skills/dh-cli-usage/references/mcp-connection-check.md
```diff
@@ -1,34 +1,34 @@
...
+
+2. When it fails again, start each server directly and read what it prints:
+
+   ```bash
+   uv run --script "<dh_scripts/>/run_backlog_server.py"
```
Jamie-BitFlight, line 12: **Two blocking stdio servers in one `bash` fence — an agent that runs the block hangs until its tool timeout.**

Both scripts end in `mcp.run()` with stdio transport (`scripts/run_backlog_server.py`, `scripts/run_sam_server.py`). The previous revision put them in a ```` ```text ```` fence; making it ```` ```bash ```` invites direct execution, and an agent pasting both lines into one Bash call blocks on the first one forever — the prose mitigation ("stop it before you start the next one") is below the fence and easy to miss.

The success criterion is also not observable from an agent's non-interactive shell: "a process that stays up and prints no error is a working server" requires watching a process that, launched with no MCP client on stdin, either blocks or exits at EOF. Neither outcome distinguishes a healthy server from a broken one.

Consider one command per fence, wrapped in the repo's own bounded runner (`uv run --script scripts/run_bounded.py --timeout-seconds 10 -- …`, per AGENTS.md gotcha 4), and state the pass/fail signal as "exits non-zero with a traceback" rather than "stays up".

---

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
```diff
@@ -965,6 +1052,73 @@ def test_cli_guide_and_connection_check_live_in_dh_cli_usage() -> None:
...
+    )
+
+    source_step = "Step 3 -- Extract context for proportional gates"
+    assert source_step in final_handoff, f"final-handoff.md does not name `{source_step}` as the {{item_ref}} source."
+
+    heading_at = skill.find(f"**{source_step}**")
+    assert heading_at != -1, (
+        f"complete-implementation/SKILL.md has no `{source_step}` step for final-handoff.md to cite."
+    )
+    assert "- `item_ref`: str (the response's opaque `reference`)" in skill, (
```
Jamie-BitFlight, line 1108: **This assertion pins the non-existent `reference` field into the guard suite.**

See the finding on `final-handoff.md:16`: `backlog_view` returns no `reference` key on either response shape. This test makes `complete-implementation/SKILL.md`'s `` - `item_ref`: str (the response's opaque `reference`) `` line mandatory, so fixing the skill to read the field that does exist (`issue`) now fails the test.

Scenario: a follow-up commit corrects Step 3 to `` - `item_ref`: str (the response's `issue`) ``, the handoff starts working again, and CI goes red on this line. Update the expected string in the same change that fixes the skill.

---

plugins/development-harness/CLAIMS-REGISTER.md
Jamie-BitFlight: Addressed in 76171326f. I narrowed the universal MCP claim and added current primary sources for pi, Cursor, and OpenCode, so the conclusion now matches the cited evidence without the earlier overstatement.

plugins/development-harness/skills/dh-cli-usage/SKILL.md
Jamie-BitFlight: Addressed in 76171326f. I removed the false positional phrase “above this body”; the fallback now uses any absolute directory supplied with the skill, returned as skill_root, or used to read SKILL.md.

plugins/development-harness/skills/complete-implementation/references/final-handoff.md
Jamie-BitFlight: Addressed in 76171326f. The caveat is now provider-neutral: it describes a provider returning a raw body instead of section-filtered content, without treating Concerns as GitHub-specific.

plugins/development-harness/docs/work-ledger/measurements/harness-cursor.md
Jamie-BitFlight: Addressed in 76171326f. The paragraph now states that this evidence concerns dh-cli-usage runtime path resolution, not Cursor MCP configuration, and keeps the unsupported absolute-root claim explicitly unestablished.

plugins/development-harness/skills/dh-cli-usage/references/mcp-connection-check.md
Jamie-BitFlight: Addressed in 76171326f. The recovery procedure is now agent-executable: bounded MCP handshakes, explicit success/failure observables, and exact stderr preservation for a blocked report.

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
Jamie-BitFlight: Addressed in 76171326f. I removed the long speculative comment and left the executable guard and names to express the contract.

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
Jamie-BitFlight: Addressed in 76171326f. I removed the redundant prose around the pinned instruction tuple; the test itself now carries the contract.

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
Jamie-BitFlight: Addressed in 76171326f. The test now resolves local links in SKILL.md, COVERAGE.md, the methodology, and the rule, and asserts both required rule globs. This protects behavior rather than old string locations.

plugins/development-harness/skills/meta-workflow-graph-refresh/references/COVERAGE.md
Jamie-BitFlight: Fixed in 76171326f. COVERAGE.md now uses ./workflow-trace-methodology.md and ../../../docs/plans/prd-workflow-extractor.md, both resolved from its relocated directory.

plugins/development-harness/skills/meta-workflow-graph-refresh/references/workflow-trace-methodology.md
Jamie-BitFlight: Fixed in 76171326f. The methodology PRD reference now resolves through ../../../docs/plans/prd-workflow-extractor.md from its relocated references directory.

plugins/development-harness/skills/complete-implementation/references/final-handoff.md
Jamie-BitFlight: Addressed in 76171326f. Final handoff now accepts item_ref from the direct work-item response issue, the plan row issue, or the slug-search result issue. The parent lookup runs before the beads-only label skip.

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
Jamie-BitFlight: Addressed in 76171326f. The regression test now includes COVERAGE.md and resolves each local Markdown/backtick link against its source file; it also checks both workflow rule globs.

plugins/development-harness/skills/dh-cli-usage/references/mcp-connection-check.md
Jamie-BitFlight: Addressed in 76171326f. Each server is diagnosed independently from the project root via FastMCP’s client handshake, under the vendored bounded runner, with --project-dir . and explicit exit semantics.

plugins/development-harness/skills/dh-cli-usage/SKILL.md
Jamie-BitFlight: Fixed in 76171326f. The fallback no longer assumes the base-directory line appears above the body, so append-after-body harnesses remain valid.

plugins/development-harness/skills/dh-cli-usage/references/mcp-connection-check.md
Jamie-BitFlight: Restored in 76171326f. The recovery starts with a session restart and scopes MCP_TIMEOUT remediation explicitly to Claude Code startup-timeout stderr.

plugins/development-harness/skills/complete-implementation/references/final-handoff.md
Jamie-BitFlight: Fixed in 76171326f. BacklogViewResponse exposes issue, not reference; complete-implementation now uses the non-empty response issue or retains the normalized input selector.

plugins/development-harness/skills/dh-cli-usage/SKILL.md
Jamie-BitFlight: Addressed in 76171326f. The fallback again permits the absolute directory used to read SKILL.md, alongside supplied-directory and skill_root sources, while retaining the fail-closed path when none is absolute.

plugins/development-harness/.claude/rules/workflow-extraction.md
Jamie-BitFlight: Fixed in 76171326f. The workflow rule now covers both docs/workflow-layers/** and skills/meta-workflow-graph-refresh/references/**.

plugins/development-harness/skills/meta-workflow-graph-refresh/SKILL.md
Jamie-BitFlight: Fixed in 76171326f with the shared link correction noted in the earlier thread; both moved documents now resolve their local references.

plugins/development-harness/skills/dh-cli-usage/references/mcp-connection-check.md
Jamie-BitFlight: Addressed in 76171326f. The commands no longer merely launch blocking stdio servers: each performs a real bounded fastmcp list handshake and treats exit 0 plus a tool list as success.

plugins/development-harness/tests/test_agent_portability_and_verdict_drift.py
Jamie-BitFlight: Fixed in 76171326f. The regression now pins the actual issue field for the direct response, plan row, and slug-search result, and rejects the nonexistent response reference wording.

Jamie-BitFlight:
Follow-up to review https://github.com/Jamie-BitFlight/claude_skills/pull/3617#pullrequestreview-5253948681

Commit 76171326f addresses all five findings: relocated links resolve and are guarded; both workflow rule globs are restored; handoff selectors use the real issue field across entry paths; and MCP diagnostics now perform separate bounded client handshakes from the project root. Focused tests and changed-file hooks pass.

Jamie-BitFlight:
Follow-up to review https://github.com/Jamie-BitFlight/claude_skills/pull/3617#pullrequestreview-5254265626

Commit 76171326f addresses all eight contract findings: issue replaces nonexistent reference; positional base-directory wording is removed; session restart and Claude-only MCP_TIMEOUT guidance are restored; the resolved SKILL.md path fallback is retained; workflow scope is restored; and the duplicate link/MCP/test findings share the same verified source fixes.

Jamie-BitFlight:
Follow-up to review https://github.com/Jamie-BitFlight/claude_skills/pull/3617#pullrequestreview-5236525601

I completed the requested claim checking and adversarial pass in commit 76171326f. The unsupported universal MCP statement was narrowed, current primary sources were added, speculative test commentary was removed, and every related runtime claim now either has an evidence source or remains explicitly unestablished.

PR #3642, opened 2026-09-18
fix(dh): make backlog_view runtime hints R7-compliant

Jamie-BitFlight:
Refs #3060 (partial: this PR does not close the issue)

## What this PR does not resolve

The #3060 acceptance criteria compared with this change:

| #3060 criterion | Status after this PR |
|---|---|
| No hint recommends retrieving everything first | Met for `_hint`. The `summary` parameter description still reads "`_hint` showing how to load full content or specific sections" (full content first). |
| The over-budget hint offers an action a narrowed caller has not taken | Met: `_usage` names `map=True`. |
| Hints name only mechanisms present in the response | Not met for the narrowed `_usage`: `map=`, `navigate=`, `head=` and `skip_tokens=` do not appear in the over-budget response. |
| Addresses in a hint are valid in that response | The narrowed `_usage` names no address. |

The narrowed hint depends on an unbounded response. #3060's grooming accepted `map=True` because `map=True` "is documented as always under 2,000 tokens". #3059 shows that claim is false: MAP has no size limit (`backlog_core/ARCHITECTURE.md`, "Resolution Index vs Map Text Split"). On a large item, a caller that follows this hint can get a response over the harness cap. The hint becomes safe for every item size when #3059 paginates MAP.

The groomed acceptance criteria 2 and 4 on #3060 name solutions and test functions. They are not observable checks, which `groom/swarm.md` requires, so #3060 needs to be groomed again before it can close.

## Problem

Two `backlog_view` hints broke R7 ("Hints must be actionable") in `plugins/development-harness/docs/agent-markdown-consumption-contract.md`:

- `_build_compact_manifest()` (`summary=True`): the first line of `_hint` told the agent to load all content with `summary=False`.
- `_build_over_budget_view()` (`summary=False`): `_usage` always told the agent to narrow with `section=`/`sections=`, even when the agent had already narrowed to one section and that section alone was still over budget.

## Change (`backlog_core/server.py` only)

- `_hint` now lists the section-narrowing call first and the full-content load second.
- `_build_over_budget_view()` takes a new keyword-only flag, `narrowed_to_single_section=False`. `backlog_view` sets it when the caller passed `section=` or `sections=` and exactly one section is left in `full_response["sections"]`. It reads `full_response` because `_filter_view_sections()` narrows `full_response` but not `result` for `sections=[...]`. When the flag is set, `_usage` names `backlog_view(selector=..., map=True)` and does not suggest narrowing again.
- The hint names no `navigate=<ordinal>`. The `[N]` numbers in `sections_index` come from enumerating a dict, so they can differ from the `OrdinalPathMapper` ordinals when two sections share a title. `map=True` needs no address.

Every existing call site and substring assertion stays the same. When the flag is not set, the `_usage` text is byte-identical to before.

## Tests

- `test_build_compact_manifest_hint_recommends_section_narrowing_first` (unit)
- `test_build_over_budget_view_narrowed_usage_advises_ordinal_map_not_section_narrowing` (unit)
- `TestReviewRoundContract::test_over_budget_narrowed_single_section_usage_advises_ordinal_map`: a request-level test that proves the call site sets the flag

Against origin/main's `server.py`, all three new tests fail. With the fix they pass. `tests/test_backlog_core_server.py` plus `backlog_core/tests/` all pass: 1151 tests. prek (ruff, ruff-format, ty) passes.

https://claude.ai/code/session_01ARwvHiWMwQy382yG5bqkAG


Jamie-BitFlight:
Replaced by #3649. That PR carries the same two commits (a90ed1e06, dc5b6d264) without 7c9203b05. That commit removed #3060 from the backlog_core/ARCHITECTURE.md debt table, and its message said "closes #3060", which a squash merge would have applied while #3060 is only partly fixed. Repo owner decision, 2026-09-19: move the work to a new branch instead of force-pushing this one.

